PR #265, opened 2017-12-21
Fix runtestplan infra paramaters

sameerawickramasekara:


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -60,6 +61,12 @@ public Deployment createInfrastructure(Infrastructure infrastructure, String inf
         awsManager.init(infrastructure);
         for (Script script : infrastructure.getScripts()) {
             if (script.getScriptType().equals(Script.ScriptType.CLOUD_FORMATION)) {
+                for (Map<String, String> stringStringMap : infrastructure.getInfraParams()) {
+                    stringStringMap.entrySet().forEach(stringStringEntry -> {
+                        script.getScriptParameters().setProperty(stringStringEntry.getKey(),
+                                stringStringEntry.getValue());
+                    });
+                }
```
kasunbg: Why is lambda forEach and for-loop mixed here. Is it possible to pick one and use that?

PR #295, opened 2018-01-03
Create dashboard WAR archive during TestGrid build.

### LINKED ISSUE #294: Add npm build of TestGrid dashboard to WAR distribution.
**Description:**
Currently, TestGrid dashboard is not built when TestGrid maven build is triggered. we need to trigger an npm build from maven and add the build folder to WAR distribution, so that latest changes are always available in the war archive.

sameerawickramasekara:
## Purpose
Resolves #294

## Goals
TestGrid dashboard build will be triggered when a TestGrid build is triggered. and a WAR archive will be generated that can be deployed on tomcat.

## Approach
'npm run build' is executed through maven exec plugin and copying the content to webapp folder is done via maven antrun plugin.

## Security checks


web/pom.xml
```diff
@@ -89,28 +87,105 @@
...
+                            </tasks>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
```
vidurananayakkara: This configuration is inherited from the parent pom and therefore we need not to explicitly specify

web/pom.xml
```diff
@@ -89,28 +87,105 @@
...
+            <plugin>
+                <artifactId>maven-war-plugin</artifactId>
+                <configuration>
+                    <warName>testgrid#v0.9</warName>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.tomcat.maven</groupId>
+                <artifactId>tomcat7-maven-plugin</artifactId>
+                <version>2.2</version>
```
vidurananayakkara: Better to define all the plugins in the plugin management section of the parent pom. Furthermore, we can add versions as properties. The purpose of this would be easing of changing the plugin/dependency version

web/pom.xml
```diff
@@ -89,28 +87,105 @@
...
-                <plugin>
-                    <groupId>org.apache.tomcat.maven</groupId>
-                    <artifactId>tomcat7-maven-plugin</artifactId>
-                    <version>2.2</version>
-                </plugin>
-            </plugins>
-        </pluginManagement>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
```
vidurananayakkara: Better to define all the plugins in the plugin management section of the parent pom. Furthermore, we can add versions as properties. The purpose of this would be easing of changing the plugin/dependency version.

The following can be added to the plugin management section of the parent pom
```xml
<plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>exec-maven-plugin</artifactId>
    <version>${org.codehaus.mojo.version}</version>
```

web/pom.xml
```diff
@@ -89,28 +87,105 @@
...
+                        <id>prepare-deploy-package</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <tasks>
+                                <copy todir="${basedir}/src/main/webapp/">
+                                    <fileset dir="${basedir}/src/main/react-dashboard/build/"
+                                             includes="**/*"/>
```
kasunbg: All newly build artifacts should go inside the target/ folder. So, `src/main/react-dashboard/build` and `src/main/webapp` may not be proper places for build artifacts.

web/pom.xml
```diff
@@ -89,28 +87,105 @@
...
+                            <goal>exec</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <executable>npm</executable>
+                            <arguments>
+                                <argument>run</argument>
+                                <argument>build</argument>
+                            </arguments>
+                            <workingDirectory>${basedir}/src/main/react-dashboard</workingDirectory>
```
kasunbg: Shall we set the workingDirectory to somewhere in the target/ folder? Is it possible?

web/src/main/react-dashboard/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "testgrid-dashboard",
   "version": "0.1.0",
   "private": true,
-  "homepage" : "http://ec2-34-238-28-168.compute-1.amazonaws.com:8080/testgrid/v0.9/",
+  "homepage": "http://ec2-34-238-28-168.compute-1.amazonaws.com:8080/testgrid/v0.9/",
```
kasunbg: We should mention about this config in the README.md.

vidurananayakkara: Is it a must to specify the hostname? What is the possibility of stating only the URI?

web/pom.xml
sameerawickramasekara: Create issue #300 to track this issue

web/pom.xml
sameerawickramasekara: tracked by issue #300

PR #299, opened 2018-01-04
Improve TestGrid Logging Structure

### LINKED ISSUE #272: The log file structure should be simpler
**Description:**
The current log file structure contains too many log files. This should be reduced and the structure should be revisited.

**Suggested Labels:**
Type/Task

**Suggested Assignees:**
AsmaJ

vidurananayakkara:
## Purpose
Resolves #272 
Simplify the logging structure

## Goals
Simplify logging structure

## Approach
image

## Release note
* Improved TestGrid logging structure

## Documentation
Included in this PR

## Security checks

## Test environment
JDK 8u151, Ubuntu 16.04


core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -130,22 +132,25 @@ public void execute() throws CommandExecutionException {
...
+            DeploymentPattern deploymentPattern = getDeploymentPattern(product,
+                    testConfig.getDeploymentPatterns().get(0));
 
-            //Set logger file path
-            setLogFilePath(product, testPlan);
+            // Generate test plan from config
+            TestPlan testPlan = generateTestPlan(deploymentPattern, testConfig, scenarioRepoDir, infraRepo);
+
+            //Set the log file path
+            LogFilePathLookup.setLogFilePath(deriveLogFilePath(product, testPlan));
```
kasunbg: Why are you modifying a static field here? This does not seem correct.

logging/src/main/java/org/wso2/testgrid/logging/plugins/LogFilePathLookup.java
```diff
@@ -0,0 +1,53 @@
...
+        return logFilePath;
+    }
+
+    /**
+     * Sets the log file path.
+     *
+     * @param logFilePath log file path
+     */
+    public static void setLogFilePath(String logFilePath) {
+        LogFilePathLookup.logFilePath = logFilePath;
```
kasunbg: We also need to instruct Jenkins pipeline to archive this log file. Please create a separate github issue to track it.

vidurananayakkara: Github issue - https://github.com/wso2-incubator/testgrid/issues/302

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: This path is accessed via the log4j framework via the log4j.xml. We do only know the log file path in runtime and therefore need to modify the path in runtime.

kasunbg: Setting the path at runtime is fine. But doing that in a static method is not correct. Anyway, this will be addressed via [1] as discussed. 

[1] https://github.com/wso2-incubator/testgrid/issues/302

PR #307, opened 2018-01-04
Generate test report for each test run in test artifact directory

### LINKED ISSUE #306: Generate report for each test plan in the test run artifacts location
**Description:**
Reports for each test run should be generated and saved in the relevant test run artefacts location

vidurananayakkara:
## Purpose
Test reports for each test plan will now be created in the relevant test artefacts location
Resolves #306 

## Approach
image

## Release note
* Save test reports for each test run in the test run artefacts location

## Security checks

### DIFF
--- a/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
@@ -18,6 +18,10 @@
 
 package org.wso2.testgrid.common.util;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.common.Deployment;
@@ -26,6 +30,8 @@ import org.wso2.testgrid.common.Host;
 import org.wso2.testgrid.common.Product;
 import org.wso2.testgrid.common.TestPlan;
 import org.wso2.testgrid.common.exception.CommandExecutionException;
+import org.wso2.testgrid.common.exception.TestGridException;
+import org.wso2.testgrid.common.exception.TestGridLoggingException;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -35,7 +41,9 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Locale;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.UUID;
 
 /**
@@ -161,17 +169,23 @@ public final class TestGridUtil {
      * @param testPlan test plan for getting the test run artifacts location
      * @return path of the test run artifacts
      */
-    public static Path getTestRunArtifactsDirectory(TestPlan testPlan) {
-        DeploymentPattern deploymentPattern = testPlan.getDeploymentPattern();
-        Product product = deploymentPattern.getProduct();
-        int testRunNumber = testPlan.getTestRunNumber();
+    public static Path getTestRunArtifactsDirectory(TestPlan testPlan) throws TestGridLoggingException {
+        try {
+            DeploymentPattern deploymentPattern = testPlan.getDeploymentPattern();
+            Product product = deploymentPattern.getProduct();
+            int testRunNumber = testPlan.getTestRunNumber();
 
-        String productDir = StringUtil.concatStrings(product.getName(), "_", product.getVersion()
-                , "_", product.getChannel());
-        String deploymentDir = deploymentPattern.getName();
-        String infraDir = getInfraParamUUID(testPlan.getInfraParameters());
+            String productDir = StringUtil.concatStrings(product.getName(), "_", product.getVersion()
+                    , "_", product.getChannel());
+            String deploymentDir = deploymentPattern.getName();
+            String infraDir = getInfraParamUUID(testPlan.getInfraParameters());
 
-        return Paths.get(productDir, deploymentDir, infraDir, String.valueOf(testRunNumber));
+            return Paths.get(productDir, deploymentDir, infraDir, String.valueOf(testRunNumber));
+        } catch (TestGridException e) {
+            throw new TestGridLoggingException(
+                    "Error in getting the test run artifacts directory location " +
+                    "([PRODUCT_NAME_VERSION_CHANNEL]/[DEPLOYMENT_PATTERN_NAME]/[INFRA_PARAM_UUID]/[TEST_RUN_NUMBER]");
+        }
     }
 
     /**
@@ -180,7 +194,27 @@ public final class TestGridUtil {
      * @param infraParams infra parameters to get the UUID
      * @return UUID specific to the infra parameters
      */
-    private static String getInfraParamUUID(String infraParams) {
-        return UUID.nameUUIDFromBytes(infraParams.getBytes(Charset.defaultCharset())).toString();
+    private static String getInfraParamUUID(String infraParams) throws TestGridException {
+        try {
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, Object> sortedMap = new TreeMap<>();
+
+            // Get sorted map from the JSON object string
+            Map<String, Object> infraParamMap = mapper
+                    .readValue(infraParams, new TypeReference<Map<String, String>>() {
+                    });
+
+            // Copy values to the tree map to get the values sorted
+            infraParamMap.forEach(sortedMap::put);
+            String stringToConvertToUUID = sortedMap.toString().toLowerCase(Locale.ENGLISH);
+            return UUID.nameUUIDFromBytes(stringToConvertToUUID.getBytes(Charset.defaultCharset())).toString();
+        } catch (JsonParseException e) {
+            throw new TestGridException(StringUtil.concatStrings("Error in parsing JSON ", infraParams), e);
+        } catch (JsonMappingException e) {
+            throw new TestGridException(StringUtil.concatStrings("Error in mapping JSON ", infraParams), e);
+        } catch (IOException e) {
+            throw new TestGridException(StringUtil.concatStrings("IO Exception when trying to map JSON ",
+                    infraParams), e);
+        }
     }
 }
--- a/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
@@ -21,7 +21,10 @@ package org.wso2.testgrid.common.util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.common.Deployment;
+import org.wso2.testgrid.common.DeploymentPattern;
 import org.wso2.testgrid.common.Host;
+import org.wso2.testgrid.common.Product;
+import org.wso2.testgrid.common.TestPlan;
 import org.wso2.testgrid.common.exception.CommandExecutionException;
 
 import java.io.BufferedReader;
@@ -153,12 +156,31 @@ public final class TestGridUtil {
     }
 
     /**
+     * Returns the directory location where the test run artifacts resides.
+     *
+     * @param testPlan test plan for getting the test run artifacts location
+     * @return path of the test run artifacts
+     */
+    public static Path getTestRunArtifactsDirectory(TestPlan testPlan) {
+        DeploymentPattern deploymentPattern = testPlan.getDeploymentPattern();
+        Product product = deploymentPattern.getProduct();
+        int testRunNumber = testPlan.getTestRunNumber();
+
+        String productDir = StringUtil.concatStrings(product.getName(), "_", product.getVersion()
+                , "_", product.getChannel());
+        String deploymentDir = deploymentPattern.getName();
+        String infraDir = getInfraParamUUID(testPlan.getInfraParameters());
+
+        return Paths.get(productDir, deploymentDir, infraDir, String.valueOf(testRunNumber));
+    }
+
+    /**
      * Returns a UUID specific to the infra parameters.
      *
      * @param infraParams infra parameters to get the UUID
      * @return UUID specific to the infra parameters
      */
-    public static String getInfraParamUUID(String infraParams) {
+    private static String getInfraParamUUID(String infraParams) {
         return UUID.nameUUIDFromBytes(infraParams.getBytes(Charset.defaultCharset())).toString();
     }
 }
--- a/core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
+++ b/core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
@@ -56,9 +56,10 @@ public class TestPlanExecutor {
      * This method executes a given {@link TestPlan}.
      *
      * @param testPlan an instance of {@link TestPlan} in which the tests should be executed
+     * @return executed test plan
      * @throws TestPlanExecutorException thrown when error on executing test plan
      */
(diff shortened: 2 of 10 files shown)


common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
```diff
@@ -152,13 +155,32 @@ public static String getTestGridHomePath() {
...
+
     /**
      * Returns a UUID specific to the infra parameters.
      *
      * @param infraParams infra parameters to get the UUID
      * @return UUID specific to the infra parameters
      */
-    public static String getInfraParamUUID(String infraParams) {
+    private static String getInfraParamUUID(String infraParams) {
         return UUID.nameUUIDFromBytes(infraParams.getBytes(Charset.defaultCharset())).toString();
```
sameerawickramasekara: In here shouldn't we check for same infra parameters in different forms,
like `ubuntu/mysql/jdk8` and `UBUNTU/MySQL/JDK8` since we have no control over these parameter values.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -56,9 +56,10 @@
      * This method executes a given {@link TestPlan}.
      *
      * @param testPlan an instance of {@link TestPlan} in which the tests should be executed
+     * @return executed test plan
      * @throws TestPlanExecutorException thrown when error on executing test plan
      */
-    public void runTestPlan(TestPlan testPlan, Infrastructure infrastructure) throws TestPlanExecutorException {
+    public TestPlan runTestPlan(TestPlan testPlan, Infrastructure infrastructure) throws TestPlanExecutorException {
```
kasunbg: #runTestPlan should not return the TestPlan back.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -144,18 +149,26 @@ public void execute() throws CommandExecutionException {
...
             //Set the log file path
-            LogFilePathLookup.setLogFilePath(deriveLogFilePath(product, testPlan));
+            LogFilePathLookup.setLogFilePath(deriveLogFilePath(testPlan));
 
             // Product, deployment pattern, test plan and test scenarios should be persisted
             testPlan = persistTestPlan(testPlan);
 
             // Execute test plan
-            executeTestPlan(testPlan, infrastructure);
+            testPlan = executeTestPlan(testPlan, infrastructure);
```
kasunbg: Returning the TestPlan object from the #executeTestPlan is not correct.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -144,18 +149,26 @@ public void execute() throws CommandExecutionException {
...
+
+            // Generate report for the test plan
+            TestReportEngine testReportEngine = new TestReportEngine();
```
kasunbg: Shall we use dependency injection (DI) here to get an instance of TestReportEngine? We shouldn't create instances unnecessarily like this nor use singletons since we have a better alternative. 

See Spring DI: http://www.vogella.com/tutorials/SpringDependencyInjection/article.html#dependencyinjection

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -144,18 +149,26 @@ public void execute() throws CommandExecutionException {
...
+            testReportEngine.generateReport(testPlan);
         } catch (IllegalArgumentException e) {
             throw new CommandExecutionException(StringUtil.concatStrings("Channel ",
                     channel, " is not defined in the available channels enum."), e);
         } catch (IOException e) {
             throw new CommandExecutionException("Error in reading file generated config file", e);
+        } catch (ReportingException e) {
+            throw new CommandExecutionException("Error in generating report for the test plan.", e);
+        } catch (TestGridException e) {
+            throw new CommandExecutionException("Error in deriving log file path.", e);
```
kasunbg: We cannot say TestGridException will be thrown only for "Error in deriving log file path"! How about we throw a specific exception from LogFilePathLookup.setLogFilePath and catch that here?

Possible signature for the exception class: `public class LogFileDeriveException extends TestGridException`

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -271,13 +284,14 @@ private TestPlan persistTestPlan(TestPlan testPlan) throws CommandExecutionExcep
      * This method triggers the execution of a {@link TestPlan}.
      *
      * @param testPlan test plan to execute
+     * @return executed test plan
      * @throws CommandExecutionException thrown when error on executing test plan
      */
-    private void executeTestPlan(TestPlan testPlan, Infrastructure infrastructure)
+    private TestPlan executeTestPlan(TestPlan testPlan, Infrastructure infrastructure)
```
kasunbg: See above.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -314,30 +322,36 @@ private String deriveLogFilePath(Product product, TestPlan testPlan) {
...
-        List<TestScenario> testScenarios = new ArrayList<>();
-        for (String name : testConfig.getScenarios()) {
-            TestScenario testScenario = new TestScenario();
-            testScenario.setName(name);
-            testScenario.setTestPlan(testPlan);
-            testScenario.setStatus(Status.PENDING);
-            testScenarios.add(testScenario);
+                                      String infraRepoDir) throws CommandExecutionException {
+        try {
+            String jsonInfraParams = new ObjectMapper().writeValueAsString(testConfig.getInfraParams().get(0));
```
kasunbg: Why do you get the first element here?

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -314,30 +322,36 @@ private String deriveLogFilePath(Product product, TestPlan testPlan) {
...
+            TestPlan testPlan = new TestPlan();
+            testPlan.setStatus(Status.PENDING);
```
kasunbg: We should take out the STATUS from TestPlan if possible. We can have the STATUS maintained in the same TestPlan table in the DB though.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -314,30 +322,36 @@ private String deriveLogFilePath(Product product, TestPlan testPlan) {
...
+                testScenario.setName(name);
+                testScenario.setTestPlan(testPlan);
+                testScenario.setStatus(Status.PENDING);
+                testScenarios.add(testScenario);
+            }
+            testPlan.setTestScenarios(testScenarios);
+            return testPlan;
+        } catch (JsonProcessingException e) {
+            throw new CommandExecutionException(StringUtil
+                    .concatStrings("Error in processing JSON ", testConfig.getInfraParams()), e);
```
kasunbg: What/Who is JSON here? :)

reporting/src/main/java/org/wso2/testgrid/reporting/TestReportEngine.java
```diff
@@ -68,20 +67,58 @@
...
+            String htmlString = renderable.render(REPORT_MUSTACHE, parsedResultMap);
+
+            // Write to HTML file
+            String testRunArtifactsDir = TestGridUtil.getTestRunArtifactsDirectory(testPlan).toString();
+            String fileName = StringUtil.concatStrings("Report", HTML_EXTENSION);
+            String testGridHome = TestGridUtil.getTestGridHomePath();
+            Path reportPath = Paths.get(testGridHome).resolve(testRunArtifactsDir).resolve(fileName);
+            writeHTMLToFile(reportPath, htmlString);
+        } catch (TestGridException e) {
+            throw new ReportingException("Error in deriving log file path.", e);
```
kasunbg: Incorrect error description.

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
```diff
@@ -75,12 +78,40 @@ public void execute() throws CommandExecutionException {
...
+     */
+    private Product getProduct(String productName, String productVersion, String channel)
+            throws CommandExecutionException {
+        try {
+            ProductUOW productUOW = new ProductUOW();
+            Product.Channel productChannel = Product.Channel.valueOf(channel);
+            return productUOW.getProduct(productName, productVersion, productChannel)
+                    .orElseThrow(() -> new CommandExecutionException(StringUtil
+                            .concatStrings("No product test plan found for product {product name: ", productName,
+                                    ", product version: ", productVersion, ", channel: ", channel, "}")));
```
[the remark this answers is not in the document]
vidurananayakkara: This will be logged in [1]. So when the exception is thrown we know what it caused and the action performed.

[1] https://github.com/vidurananayakkara/wso2-test-grid/blob/a508c35a2f3d928681b8b239c4ea39946cb2a592/core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java#L85

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
```diff
@@ -75,12 +78,40 @@ public void execute() throws CommandExecutionException {
...
+        } catch (IllegalArgumentException e) {
+            throw new CommandExecutionException(StringUtil.concatStrings("Channel ", channel,
+                    " not found in channels enum."));
```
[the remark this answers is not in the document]
vidurananayakkara: Channels enum values will be printed along to identify which values are applicable.

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
```diff
@@ -75,12 +78,40 @@ public void execute() throws CommandExecutionException {
...
+        } catch (TestGridDAOException e) {
+            throw new CommandExecutionException("Error in instantiating DB transaction.", e);
```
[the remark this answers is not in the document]
vidurananayakkara: Exception message modified.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: Created a git issue to track this - https://github.com/wso2-incubator/testgrid/issues/322

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: In this case, `TestGridException` will be only thrown for this reason. Furthermore, the error actually occurs when creating a UUID using infra params (not due to actually setting the log file path). Hence it is TestGrid core related than logging framework related.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: When the test plan command is run we are only considering a set of infra parameters (only a set of infra params can be there for the test plan). The array which comes here only has 1 element.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: Fixed. Statuses will be set where appropriate.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: Refactored the error message

reporting/src/main/java/org/wso2/testgrid/reporting/TestReportEngine.java
vidurananayakkara: Error message refactored.

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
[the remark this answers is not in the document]
kasunbg: Ah.. I do not see remedial information in there though. May be I'm missing something.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
kasunbg: What will when this code evolves? What if someone adds a new logic that throws the TestGridException? I do not think he will notice to update the log.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
kasunbg: As I understand, you are saying that testConfig.getInfraParams list always has a single element? Can it be empty?
Can it be more than one? If not, is there a particular reason to have a List as the return type?

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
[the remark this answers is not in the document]
kasunbg: I meant from the end-user perspective, what is a channels enum? We can be more descriptive here.

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
```diff
@@ -75,12 +78,42 @@ public void execute() throws CommandExecutionException {
...
+                    .orElseThrow(() -> new CommandExecutionException(StringUtil
+                            .concatStrings("No product test plan found for product {product name: ", productName,
+                                    ", product version: ", productVersion, ", channel: ", channel, "}")));
+        } catch (IllegalArgumentException e) {
+            throw new CommandExecutionException(StringUtil.concatStrings("Channel ", channel,
+                    " not found in channels enum - [ ", Product.Channel.values(), " ]"));
+        } catch (TestGridDAOException e) {
+            throw new CommandExecutionException(StringUtil
+                    .concatStrings("Error in retrieving Product {product name: ", productName,
+                            ", product version: ", productVersion, ", channel: ", channel, "} from the Database."), e);
```
kasunbg: Missing remedial actions.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: Added a custom exception `TestGridLoggingException`

core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java
vidurananayakkara: This will be logged in [1]. So when the exception is thrown we know what it caused and the action performed.

[1] https://github.com/vidurananayakkara/wso2-test-grid/blob/a508c35a2f3d928681b8b239c4ea39946cb2a592/core/src/main/java/org/wso2/testgrid/core/command/GenerateReportCommand.java#L85

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
vidurananayakkara: If this is empty the `jsonInfraParams` will be an empty JSON string. Our input is a YAML file containing a list of infra parameters. We create test plans out of this YAML file. The reason why we are getting a list here is since we have used the same class to read the generated test plans. This way a test plan that is generated out of this process can be fed as an input by the user. These complexities will move away if we ever decide to store the test plans in the DB rather than generating YAML files.

kasunbg: This is something we need to fix. But not an issue for this PR.

PR #364, opened 2018-01-18
Persist the Build API request payload as a config file.

### LINKED ISSUE #355: Persisting the Build API request payload as a config file.
**Description:**
The payload which comes with the Build API (API used to create test plans) will be saved as a config file to be referred in the pipeline script.

**Related Issues:**
#328

pasindujw:
## Purpose
The payload which comes with the Build API (API used to create test plans) will be saved as a config file to be referred in the pipeline script.
Fixes #355 

## Approach
 Using FasterXML jackson, the Java object is parsed and written to a YAML file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -338,11 +338,6 @@
                 <artifactId>jackson-annotations</artifactId>
                 <version>${com.fasterxml.jackson.version}</version>
             </dependency>
-            <dependency>
-                <groupId>com.fasterxml.jackson.dataformat</groupId>
-                <artifactId>jackson-dataformat-yaml</artifactId>
-                <version>${com.fasterxml.jackson.version}</version>
-            </dependency>
 
             <!-- GlassFish dependencies-->
             <dependency>
@@ -524,7 +519,7 @@
         <awaitility.version>3.0.0</awaitility.version>
         <guava.version>19.0</guava.version>
         <snakeyaml.version>1.19</snakeyaml.version>
-        <com.fasterxml.jackson.version>2.9.3</com.fasterxml.jackson.version>
+        <com.fasterxml.jackson.version>2.8.10</com.fasterxml.jackson.version>
         <apache.httpcomponents.fluent.version>5.0-alpha3</apache.httpcomponents.fluent.version>
         <!-- Logging dependencies -->
         <slf4j.api.version>1.7.25</slf4j.api.version>
--- a/pom.xml
+++ b/pom.xml
@@ -338,6 +338,11 @@
                 <artifactId>jackson-annotations</artifactId>
                 <version>${com.fasterxml.jackson.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.dataformat</groupId>
+                <artifactId>jackson-dataformat-yaml</artifactId>
+                <version>${com.fasterxml.jackson.version}</version>
+            </dependency>
 
             <!-- GlassFish dependencies-->
             <dependency>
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -143,4 +148,39 @@ public Response createTestPlan(TestPlanRequest testPlanRequest) {
...
+     *              1.If file directory (which is not existing) can not be created.
+     *              2.If IOException occured while writing YAML file.
+     *              3.If the YAML file already exists.
+     */
+    private void persistAsYamlFile(TestPlanRequest testPlanRequest) throws TestGridException {
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        String path = Constants.TESTPLANS_DIR + testPlanRequest.getTestPlanName();
+
+        File directory = new File(path);
+        File yamlFile = new File(path + File.separator + testPlanRequest.getTestPlanName() + ".yml");
```
kasunbg: At WSO2, we use .yaml as the extension for yaml files.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -143,4 +148,39 @@ public Response createTestPlan(TestPlanRequest testPlanRequest) {
...
+
+        if (!directory.exists()) {
+            if (!directory.mkdir()) {
```
kasunbg: This will only create the child folder.. If you want the parent folders also created, then we need to specifically say so. ie. the equivalent of UNIX 'mkdir -p' argument.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
pasindujw: Fixed.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
pasindujw: Fixed by moving to `java.nio.file.Files.createDirectories ` which creates all non existing directories in the given path.

pom.xml
```diff
@@ -338,6 +338,11 @@
                 <artifactId>jackson-annotations</artifactId>
                 <version>${com.fasterxml.jackson.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.dataformat</groupId>
+                <artifactId>jackson-dataformat-yaml</artifactId>
+                <version>${com.fasterxml.jackson.version}</version>
+            </dependency>
```
yasassri: I believe we can use snake-yaml instead of this. Shall we use that?

pasindujw: Moved into snake yaml.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -143,4 +150,43 @@ public Response createTestPlan(TestPlanRequest testPlanRequest) {
...
+    /**
+     * Save the {@link TestPlanRequest} object as a YAML file.
+     * @param testPlanRequest TestPlanRequest needs to be saved.
+     * @throws TestGridException thrown when error occurred;
+     *              1.If file directory (which is not existing) can not be created.
+     *              2.If IOException occured while writing YAML file.
+     *              3.If the YAML file already exists.
+     */
+    private void persistAsYamlFile(TestPlanRequest testPlanRequest) throws TestGridException {
+        java.nio.file.Path path = Paths.get(Constants.TESTPLANS_DIR , testPlanRequest.getTestPlanName());
```
lasanthaDLPDS: Shall we import java.nio.file.Path package and use. Because it will make the code more readable.

PR #394, opened 2018-01-23
TESTGRID_HOME should default to ~/.testgrid if the env variable is not found

### LINKED ISSUE #393: TESTGRID_HOME should default to ~/.testgrid if the env variable is not set.
**Description:**
We shouldn't just fail if the TESTGRID_HOME env var is not found. Instead, we should default to an appropriate location such as ~/.testgrid.

**Suggested Labels:**
N/A

**Suggested Assignees:**
N/A

**Affected Product Version:**
0.9.0-m9

**Steps to reproduce:**

**Related Issues:**

kasunbg:
## Purpose
We shouldn't just fail if the TESTGRID_HOME env var is not found. Instead, we should default to an appropriate location such as ~/.testgrid.

Resolves #393

## Goals
Better user experience

## Approach
Defaults to ~/.testgrid

## User stories
N/A

## Release note
N/A

## Documentation
Default testgrid_home is set to ~/.testgrid.

## Training
N/A

## Certification
N/A

## Marketing
N/A
## Automation tests
N/A
## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

### DIFF
--- a/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
@@ -38,12 +38,15 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.UUID;
+
+import static org.wso2.testgrid.common.TestGridConstants.DEFAULT_TESTGRID_HOME;
 
 /**
  * This Util class holds the common utility methods.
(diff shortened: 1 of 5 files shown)


common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
```diff
@@ -158,7 +160,14 @@ public static String executeCommand(String command, File workingDirectory, Deplo
      */
     public static String getTestGridHomePath() {
         String testGridHome = EnvironmentUtil.getSystemVariableValue(TESTGRID_HOME_ENV);
-        Path testGridHomePath = Paths.get(testGridHome);
+        Path testGridHomePath;
+        if (testGridHome == null) {
+            logger.warn("TESTGRID_HOME environment variable not set. Defaulting to ~/.testgrid.");
+            testGridHomePath = DEFAULT_TESTGRID_HOME;
```
vidurananayakkara: What if the default directory is not available. IMO we should create a directory if not exists. WDYT?

common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
```diff
@@ -158,7 +160,14 @@ public static String executeCommand(String command, File workingDirectory, Deplo
      */
     public static String getTestGridHomePath() {
         String testGridHome = EnvironmentUtil.getSystemVariableValue(TESTGRID_HOME_ENV);
-        Path testGridHomePath = Paths.get(testGridHome);
+        Path testGridHomePath;
+        if (testGridHome == null) {
```
yasassri: Since we are doing a null check here, IMO we don't need to null-check the path again in Main#49.

common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
kasunbg: Then, it will get created via https://github.com/wso2/testgrid/blob/master/core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java#L298

But, it makes sense to create the directory here itself. Added!

common/src/main/java/org/wso2/testgrid/common/util/TestGridUtil.java
kasunbg: Done..

PR #457, opened 2018-02-02
Randomize the cloudformation script name at runtime 

### LINKED ISSUE #456: Cannot run two cloudformation-based test plans in parallel
**Description:**
Cannot run two cloudformation-based test plans in parallel. There will be a naming conflict if both cloudformation scripts tried to create stacks with same name.

We need to randomize the strings in order to fix this.

**Affected Product Version:**
0.9.0-m11

**Steps to reproduce:**
See above.

**Related Issues:**
N/A

kasunbg:
## Purpose
Randomize the cloudformation script name at runtime  to support running multiple cloudformation scripts in parallel. 

Resolves #456

## Goals
Support scaling.

## Approach
Randomize the script name under the test-plan's infrastructure section.

### DIFF
--- a/common/src/main/java/org/wso2/testgrid/common/util/StringUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/StringUtil.java
@@ -17,6 +17,8 @@
  */
 package org.wso2.testgrid.common.util;
 
+import java.util.Random;
+
 /**
  * Utility class to handle {@link String} related operations.
  *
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/wso2/testgrid/common/util/StringUtil.java
```diff
@@ -47,4 +49,15 @@ public static String concatStrings(Object... objects) {
         }
         return stringBuilder.toString();
     }
+
+    public static String generateRandomString(int characterCount) {
```
vidurananayakkara: Shall we add a java doc comment here describing the purpose of this method for future reference?

kasunbg: Added..

PR #470, opened 2018-02-07
Use stack ID instead of stack name for describing stacks

### LINKED ISSUE #469: DescribeStacksRequest should use the CloudFormation stackId instead of its name.
**Description:**
Using the stackId is the only guaranteed way to make sure we get the stack that actually belong to this particular test-run.                                                                                                                                               

**Affected Product Version:**
 0.9.0-m12                                                                                                                      

  N/A                                                                                                      
**Steps to reproduce:**
  See above                                                                                                                                     
**Related Issues:**
 N/A

kasunbg:
## Purpose
Use stack ID instead of stack name for describing stacks.
Fixes #469.

## Goals
Using the stackId is the only guaranteed way to make sure we get the stack that actually belong to this particular test-run.


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/aws/AWSManager.java
```diff
@@ -132,7 +136,7 @@ public Deployment createInfrastructure(Script script, String infraRepoDir) throw
             logger.info(StringUtil.concatStrings("Stack : ", cloudFormationName, " creation completed !"));
```
pasindujw: Do we need to use stack.getStackId() here in the log?

PR #489, opened 2018-02-15
Fix failing persistence for long failure massages in test cases

### LINKED ISSUE #451: JMeter Responses are too large for database
**Description:**
During the Jmeter execution, we save the requests and responses in the database. however, during the selenium tests, these responses are too large to be saved (  HTML pages ). We need to handle this issue in a suitable manner.
For more info [1]
[1] - https://testgrid-live-dev.private.wso2.com/job/wso2is-5.4.1-LTS/16/consoleFull

**Suggested Labels:**
Severity/Major
Priority / High
Type/Bug

**Affected Product Version:**
v0.9.0-m-11

azinneera:
## Security checks

### DIFF
--- a/automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterResultCollector.java
+++ b/automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterResultCollector.java
@@ -60,11 +60,7 @@ public class JMeterResultCollector extends ResultCollector {
     public void sampleOccurred(SampleEvent sampleEvent) {
             super.sampleOccurred(sampleEvent);
             SampleResult result = sampleEvent.getResult();
-            String samplerData = result.getSamplerData();
 
-            if (samplerData.length() > FAILURE_MESSAGE_LIMIT) {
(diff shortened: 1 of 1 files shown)


vidurananayakkara:
@azinneera Having the entire failure message persisted to DB was required since we didn't have traceability via logs. IMO now, we should persist a simple failure message. The user may have to go through the log file in order to dig deep into the problem (The construction of the failure message will remain as it is to be logged). This is considering the fact that a report is for analysis of the product and the log is for troubleshooting. WDYT?

azinneera:
@vidurananayakkara We did it because most of the failure messages does not exceed the limit and therefore, for small messages, the cause can be identified in the report itself. The user will need to go through the log file only for huge failure messages. WDYT?

kasunbg:
Is there a way to get a simpler failure message that describe what the actual failure is? If yes, then +1 to do that.

Otherwise, we have to show a truncated failure message that is generated from response-data and sampler-data. We can fine-tune the truncated message size (FAILURE_MESSAGE_LIMIT) which is set to 5000 right now.

yasassri:
For now, we can resolve this issue, maybe we can revisit this later.

PR #533, opened 2018-02-26
Add constant to testgrid-web context and use it in js files

### LINKED ISSUE #372: Web app context should not be hard-coded
**Description:**
Web app context (`'/testgrid/v0.9/`) is currently hard-coded. This should be defined in a single location.

harshanL:
## Security checks


web/src/main/react-dashboard/src/constants.js
```diff
@@ -3,3 +3,4 @@ export const ERROR = "ERROR";
 export const SUCCESS = "SUCCESS";
 export const RUNNING = "RUNNING";
 export const PENDING = "PENDING";
+export const TESTGRID_CONTEXT = "/testgrid/v0.9";
```
kasunbg: This is the old context. New context is /testgrid/dashboard/. We already have a constant for this I think. 
cc @sameerawickramasekara

harshanL: I think you are referring to [1]. I've checked with that context. But some parts of the UI was not loaded properly. Because of that, I've decided to go with this context. Will check it again.

[1]. https://github.com/wso2/testgrid/blob/master/web/src/main/react-dashboard/src/components/ProductStatusView.js#L39

kasunbg: Approved the changes.

PR #539, opened 2018-03-04
Update the docs to reflect the new testgrid.yaml file

kasunbg:
## Purpose
Update the docs to reflect the new testgrid.yaml file.

### DIFF
--- a/docs/Infrastructure-Deployment-Scenarios-Repository-Structure.md
+++ b/docs/Infrastructure-Deployment-Scenarios-Repository-Structure.md
@@ -1,32 +1,51 @@
-# Infrastructure / Deployment / and Scenarios Repository Structure - External consumers' view
+# Infrastructure / Deployment / and Scenarios Repository Structure - Testgrid's perspective
 
-This document describes the external consumer's view of the above mentioned repositories. Testgrid is one such consumer.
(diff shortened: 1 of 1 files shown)


docs/Infrastructure-Deployment-Scenarios-Repository-Structure.md
```diff
@@ -1,162 +1,172 @@
...
+          phase: CREATE
 ```
 
-Note: the metadata under _inputsForInfrastructurefile_ element are fed into
-the infrastructure provisioning script. This needs to be done because even though
-the infrastructure and WSO2 product deployment are orthogonal, there is some-level
-of coupling between the two. One good example is the number of VM machines required
-for a given deployment pattern. But, this should not, however, specify any
-infrastructure _provider_ specific parameters.
+<!-- Note: the metadata under _inputsForInfrastructurefile_ element are fed into -->
```
pasindujw: By making these into comments, these lines are invisible when viewing this file.
Is that the intention or do we need to show these as "comments" in the file? :)

kasunbg: Yes. This feature is no longer there. This will only come in future. Didn't want to expose this without really validating the usecase.

PR #549, opened 2018-03-07
Fix for unwanted URL caching in TestGrid Login page

### LINKED ISSUE #546: Login page click "here" link not works other than for a first time user.
**Description:**
When a session in validated, user is redirected to TestGrid login page where there is  a link to go to Identity provider's login page. But the link is not working.
This only happens for a user who has already logged in. Seems like service-worker.js which registered is occurring the problem.

**Suggested Assignees:**
pasindujw

**Steps to reproduce:**
1. Log into TestGrid.
2. Wait till session invalidates.
3. Try TestGrid page (it will redirected to TestGrid Login page).
4. Click on the link in TestGrid login page.

pasindujw:
**Purpose**
Currently TestGrid dashboard is registering a service-worker which can be used to handle offline and caching in the app. But ATM it uses a generic react service-worker and due to that, it is unable to add caching exclusions for URLs, etc. This will remove registering a 'generic' service-worker, and in future, a TestGrid-owned service-worker needs to be written and registered.
Fixes #546 
This makes the home page loading faster by reducing data amount as well. (As per current homepage content: earlier more than 1.0MB. Taken down to less than 100KB.)

**Security checks**


kasunbg:
@pasindujw we will need to add this support back in a later version? If so, please create another ticket.

pasindujw:
@kasunbg we might not needing it in close future. However its good feature to add to TestGrid roadmap.
Created a ticket: #551

PR #550, opened 2018-03-07
Improve testgrid to honor the  return status of infra provisioning script

### LINKED ISSUE #337: TestGrid need to honor return status code of infra/deploy scripts.
**Description:**
After the infrastructure creation is executed, if infrastructure creation fails the execution flow should abort, if deployment scripts fail, scenario execution should abort. We need to support this.

**Suggested Labels:**
Bug

azinneera:
**Purpose**
Contains improvements related to honouring the return status of infra provisioning script.
Resolves #337 

**Goals**
TestGrid should handle the infra creation failure gracefully when shell scripts are used.

**Approach**
Improve ShellScriptProvider#provision to consider for the boolean value returned by the ShellExecutor#executeCommand when returning the InfraProvisionResult

**Security checks**

### DIFF
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/ShellScriptProvider.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/ShellScriptProvider.java
@@ -82,11 +82,12 @@ public class ShellScriptProvider implements InfrastructureProvider {
             String parameterString = TestGridUtil.getParameterString(null, inputParameters);
             ShellExecutor executor = new ShellExecutor(null);
             InfrastructureProvisionResult result = new InfrastructureProvisionResult();
-            if (!executor.executeCommand("bash " + Paths
-                    .get(testPlanLocation, createScript.getFile()) + " " + parameterString)) {
-                throw new TestGridInfrastructureException(
-                        "Error occurred while executing the infra-provision script. " +
-                                "Script returned a non-zero status code.");
+            int exitCode = executor.executeCommand("bash " + Paths
+                    .get(testPlanLocation, createScript.getFile()) + " " + parameterString);
+            if (exitCode > 0) {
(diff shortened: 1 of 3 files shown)


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/ShellScriptProvider.java
```diff
@@ -81,9 +81,13 @@ public InfrastructureProvisionResult provision(TestPlan testPlan)
             inputParameters.setProperty(OUTPUT_DIR, workspace);
             String parameterString = TestGridUtil.getParameterString(null, inputParameters);
             ShellExecutor executor = new ShellExecutor(null);
-            executor.executeCommand("bash " + Paths
-                    .get(testPlanLocation, createScript.getFile()) + " " + parameterString);
             InfrastructureProvisionResult result = new InfrastructureProvisionResult();
+            if (!executor.executeCommand("bash " + Paths
```
pasindujw: Instead of checking boolean true/false, will it be better to throw error from executor.executeCommand function and catch it here?

harshanL: @Kasun: In that case, the caller of the ShellExecutor have to possess the knowledge of shell return statuses. For example, it has to know non-zero means a failure. IMO, we should handle all the shell script related logic inside the ShellExecutor and return a standard outcome to the caller.

harshanL: +1

kasunbg: +1. We do not log the exit code anywhere as well.

azinneera: Changes made in 4991909

Issue #552, opened 2018-03-08
Persist infrastructureConfig details in the database along with the TestPlan


harshanL:
**Description:**
Currently, testgrid does not persist infrastructureConfig details in the database. We have to stick to test-plan.yaml file due to this reason as we need infrastructureConfig information during testplan execution phase.

**Affected Product Version:**
0.9.0-m16

**OS, DB, other environment details and versions:**    
N/A

**Steps to reproduce:**
N/A

**Related Issues:**
#514

kasunbg:
I believe this is implemented. We capture the infra parameters of a given test-plan via the  'infra_parameters' column of the test_plan table.

Issue #431, opened 2018-01-30
[Dashboard] Handle UI when no test-runs of a given job were run.


kasunbg:
**Description:**
  See the below to see what I'm trying to say: :) 
![image](https://user-images.githubusercontent.com/936037/35567198-349dfc98-05ea-11e8-92ae-8ef157ec2ed2.png)

As you see, the status is green even though there has been no test-runs for this job.                                                                                                                                              

**Affected Product Version:**
 0.9.0-m10                                                                                                                       

**OS, DB, other environment details and versions:**
  N/A                                                                                                      
**Steps to reproduce:**
  See above                                                                                                                                     
**Related Issues:**
 N/A

pasindujw:
According to the dashboard implementation (Refer:
 [https://github.com/wso2/testgrid/blob/master/web/src/main/react-dashboard/src/components/SingleRecord.js](url) ), this should be automatically fixed with #514

harshanL:
Closed with PR #555.

PR #555, opened 2018-03-08
Persist TestPlan to database at generate-TestPlan step

### LINKED ISSUE #514: GenerateTestPlanCommand should set the PENDING state in database
**Description:**
Right now, our UI only works in the happy path. Whenever a build gets broken for some issue in testgrid or in repo scripts, the UI goes haywire. It even fails to let you download the log file.

Therefore, we need to set the PENDING state into the database before attempting to do the actual task. After the task completion, if an exception is thrown, then we need to set FAILED state into the database.

**Affected Product Version:**
0.9.0-m14

N/A

**Steps to reproduce:**

**Related Issues:**
N/A

harshanL:
**Purpose**
To persist testplan at the generate-test-plans step instead of run-testplan step. This will also resolve #514 and #431.

**Goals**
TestGrid should persist test-plan information to DB at the generate-testplan step. 

**Approach**
Persisted testplan to the database when executing generate-testplan command and saved the testplan.yaml with the testplan-id. When running run-test-plan command, it will fetch the appropriate testplan object from database using the id.

**Security checks**

**Related Issues**
#431, #552

### DIFF
--- a/core/src/test/java/org/wso2/testgrid/core/command/GenerateTestPlanCommandTest.java
+++ b/core/src/test/java/org/wso2/testgrid/core/command/GenerateTestPlanCommandTest.java
@@ -33,13 +33,17 @@ import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import org.wso2.testgrid.common.DeploymentPattern;
 import org.wso2.testgrid.common.Product;
 import org.wso2.testgrid.common.TestGridConstants;
+import org.wso2.testgrid.common.TestPlan;
 import org.wso2.testgrid.common.infrastructure.DefaultInfrastructureTypes;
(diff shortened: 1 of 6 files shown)


core/src/test/java/org/wso2/testgrid/core/command/GenerateTestPlanCommandTest.java
```diff
@@ -79,6 +91,14 @@ public void init() throws Exception {
         int random = new Random().nextInt(100);
         product.setName("wso2-" + random);
 
+        this.testPlan = new TestPlan();
+        testPlan.setId("abc");
```
kasunbg: Shall we use consistent Test plan IDs? I see that we used TP_1 in the other test case.

harshanL: @kasunbg : Haven't add a constant since it has not been used more than once. Do we need to?

PR #566, opened 2018-03-13
Treat InfrastructureParameter properties as first-class Infrastructure Parameters

kasunbg:
**Purpose**
Treat InfrastructureParameter properties as first-class Infrastructure Parameter instances at runtime

**Goals**
This is useful when two or more infrastructure parameter belong together.
Classic example is the database and database version. These two must be coupled together
and added as a single Infrastructure Parameter.

**Approach**

So, we will add a wrapper InfrastructureParameter called 'Database', and add the database engine
(DBEngine) and version (DBEngineVersion) as its properties.

The properties column in InfrastructureParameter table must conform the .properties file format.

**Automation tests**
Yes.
**Security checks**

### DIFF
--- a/core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
+++ b/core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
@@ -27,10 +27,15 @@ import org.wso2.testgrid.common.infrastructure.InfrastructureValueSet;
 import org.wso2.testgrid.dao.TestGridDAOException;
 import org.wso2.testgrid.dao.uow.InfrastructureParameterUOW;
 
+import java.io.IOException;
+import java.io.StringReader;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
+import java.util.Properties;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * This class provides list of infrastructure combinations. Scenario tests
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -85,10 +94,41 @@
...
                 finalInfrastructureCombinations.add(clone);
+                addSubpropertiesAsInfrastructureParameters(clone, value);
             }
         }
 
         return finalInfrastructureCombinations;
     }
 
+    /**
+     * todo.
```
pasindujw: Missing method comment.

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -85,10 +94,41 @@
...
+    private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombination combination,
+            InfrastructureParameter origInfraParameter) {
+        try {
+            Properties properties = new Properties();
+            properties.load(new StringReader(origInfraParameter.getProperties()));
+
+            for (Map.Entry entry : properties.entrySet()) {
+                String name = (String) entry.getValue();
+                String type = (String) entry.getKey();
+                String regex = "[\\w,-\\.@:]*";
```
harshanL: It would be better to add a comment which describes the purpose of the regex.

PR #569, opened 2018-03-13
Fix build failure caused by the older version of an artifact

### LINKED ISSUE #568: testgrid build fails in jenkins due to a dependency download failed
**Description:**
An older version of org.eclipse.persistence.core is being specified in the parent pom. This no longer exists.

[WARNING] The POM for org.eclipse.persistence:org.eclipse.persistence.core:jar:2.2.0 is missing, no dependency information available

azinneera:
**Purpose**
Fix build failure
Resolves #568

**Goals**
Fix build failure

**Approach**
Changing the artifact version to latest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -487,7 +487,7 @@
             <dependency>
                 <groupId>org.eclipse.persistence</groupId>
                 <artifactId>org.eclipse.persistence.core</artifactId>
-                <version>${javax.persistence.version}</version>
+                <version>${eclipse.persistence.core.version}</version>
             </dependency>
             <!--FindBugs dependencies-->
             <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -561,6 +561,7 @@
         <com.h2database.version>1.4.196</com.h2database.version>
         <!-- DAO dependencies -->
         <javax.persistence.version>2.2.0</javax.persistence.version>
```
kasunbg: shall we get rid of javax.persistence.version property?

harshanL: +1

azinneera: The latest version of [javax.persistence](http://repo.maven.apache.org/maven2/org/eclipse/persistence/javax.persistence/) is 2.2.0 and [for org.eclipse.persistence.core](http://repo.maven.apache.org/maven2/org/eclipse/persistence/org.eclipse.persistence.core/) is 2.7.1
Hence we need to have both.

PR #583, opened 2018-03-19
Improve states and introduce ResolveStatusesCommand

### LINKED ISSUE #581: The states in testgrid should be more accurate
**Description:**
There are 4 states at the moment which are PENDING, RUNNING, SUCCESS and FAIL. These are not accurate when displaying results. We are using the FAIL state for failures in pipeline and testgrid bugs as well. SUCCESS and FAIL should be reserved for any failures in the test cases due to a bug in the product. Other failures should be indicated with different labels

Related Issues: #573 

**Suggested Labels:**
Type/Improvement

**Suggested Assignees:**
AsmaJ

azinneera:
**Purpose**
Introduces ResolveStatusesCommand and contains changes related to setting proper statuses for test-plans and test scenarios.
Resolves #581 

**Goals**
Rework invalid statuses and make them more accurate.

**Security checks**

### DIFF
--- a/automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterExecutor.java
+++ b/automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterExecutor.java
@@ -30,6 +30,8 @@ import org.wso2.testgrid.automation.TestAutomationException;
 import org.wso2.testgrid.common.DeploymentCreationResult;
 import org.wso2.testgrid.common.Host;
 import org.wso2.testgrid.common.Port;
+import org.wso2.testgrid.common.Status;
+import org.wso2.testgrid.common.TestCase;
 import org.wso2.testgrid.common.TestScenario;
 import org.wso2.testgrid.common.util.StringUtil;
 import org.wso2.testgrid.common.util.TestGridUtil;
@@ -120,6 +122,18 @@ public class JMeterExecutor extends TestExecutor {
 
         //Persist test scenario
         try {
+            if (testScenario.getTestCases().size() == 0) {
+                testScenario.setStatus(Status.ERROR);
+            } else {
+                for (TestCase testCase : testScenario.getTestCases()) {
+                    if (!testCase.isSuccess()) {
+                        testScenario.setStatus(Status.FAIL);
+                        break;
+                    } else {
+                        testScenario.setStatus(Status.SUCCESS);
+                    }
+                }
+            }
             testScenarioUOW.persistTestScenario(testScenario);
             if (logger.isDebugEnabled()) {
                 logger.debug(StringUtil.concatStrings(
--- a/common/src/main/java/org/wso2/testgrid/common/Status.java
+++ b/common/src/main/java/org/wso2/testgrid/common/Status.java
@@ -47,7 +47,7 @@ public enum Status {
     /**
      * Entity execution didn't run.
      */
-    DIDNT_RUN("DIDN'T RUN"),
+    DID_NOT_RUN("DID_NOT_RUN"),
 
     /**
      * Entity execution caused an error.
--- a/common/src/main/java/org/wso2/testgrid/common/Status.java
+++ b/common/src/main/java/org/wso2/testgrid/common/Status.java
@@ -42,7 +42,22 @@ public enum Status {
     /**
      * Entity execution is running.
      */
-    RUNNING("RUNNING");
+    RUNNING("RUNNING"),
+
+    /**
+     * Entity execution didn't run.
+     */
+    DIDNT_RUN("DIDN'T RUN"),
+
+    /**
+     * Entity execution caused an error.
+     */
+    ERROR("ERROR"),
+
+    /**
+     * Entity execution is incomplete.
+     */
+    INCOMPLETE("INCOMPLETE");
 
     private final String status;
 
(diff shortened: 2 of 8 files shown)


common/src/main/java/org/wso2/testgrid/common/Status.java
```diff
@@ -42,7 +42,22 @@
     /**
      * Entity execution is running.
      */
-    RUNNING("RUNNING");
+    RUNNING("RUNNING"),
+
+    /**
+     * Entity execution didn't run.
+     */
+    DIDNT_RUN("DIDN'T RUN"),
```
harshanL: Lets use "DID_NOT_RUN" instead of the short form.

azinneera: Fixed in 3b69fff

automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterExecutor.java
```diff
@@ -120,6 +122,18 @@ public void execute(String script, DeploymentCreationResult deploymentCreationRe
 
         //Persist test scenario
         try {
+            if (testScenario.getTestCases().size() == 0) {
+                testScenario.setStatus(Status.ERROR);
```
kasunbg: This is a DID_NOT_RUN?

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -251,10 +264,22 @@ private void releaseInfrastructure(TestPlan testPlan,
...
-    private void setTestPlanStatus(TestPlan testPlan) {
+    private void persistTestPlanStatus(TestPlan testPlan) {
         try {
-            Status status = isFailedTestScenariosExist(testPlan) ? Status.FAIL : Status.SUCCESS;
-            setTestPlanStatus(testPlan, status);
+            boolean isSuccess = true;
+            for (TestScenario testScenario : testPlan.getTestScenarios()) {
+                if (testScenario.getStatus() != Status.SUCCESS) {
+                    isSuccess = false;
+                    if (testScenario.getStatus() == Status.FAIL) {
```
kasunbg: What about other statuses of testScenarios such as DID_NOT_RUN and ERROR? How about:

```
If there is at least one FAIL, the overall status is  --> FAIL
Else, if there is at least one ERROR, the overall status is  --> ERROR
Else, the overall status is --> SUCCESS.
```
What about 'DID_NOT_RUN' status then? We have to think abt what the overall status should be in that case. The logic in this comment sets the overall status to SUCCESS for that case as well currently.

core/src/main/java/org/wso2/testgrid/core/command/CommandHandler.java
```diff
@@ -56,6 +56,8 @@
                                      impl = GenerateTestPlanCommand.class),
                          @SubCommand(name = "run-testplan",
                                      impl = RunTestPlanCommand.class),
+                         @SubCommand(name = "resolve-statuses",
+                                     impl = ResolveStatusesCommand.class),
```
kasunbg: How about **`finalize-run-testplan`** instead?

Rationale: Whenever an exception occurred in the middle of run-testplan pipeline stage, then we need to finalize that partial test-run. Hence, if an exception occurred, then we need to run _finalize-run-testplan_.

kasunbg: We need to call this command from the Jenkins pipeline exception path, right?

automation/src/main/java/org/wso2/testgrid/automation/executor/JMeterExecutor.java
azinneera: At this point, the scenario has already entered the RUNNING state and JMeter execution has started. So IMO, we should not set it as DID_NOT_RUN. DID_NOT_RUN should be used for scenarios which have not run at all (e.g. due to failure in pipeline) and that will be handled by the the ResolveStatuses command in the post block. WDYT?

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
azinneera: ERROR status is already set when scenario execution fails which is caught in TestPlanExecutor#runScenarioTests. Therefore, at this point, the ERROR status is already set if there exist any failed tests.

So additional checking of failed tests and success would suffice. WDYT?

core/src/main/java/org/wso2/testgrid/core/command/CommandHandler.java
azinneera: We have added it to the post block. Which one would be better?

azinneera: Created #602 to track this.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
azinneera: Created #610 to resolve this

core/src/main/java/org/wso2/testgrid/core/command/CommandHandler.java
kasunbg: But, then we have to wait until all testplans to complete before this is run. In the meanwhile, the Dashboard will be in an erroneous state.

I think each stage is already aware of its testplan-id. It is present in the test-plan yaml. Since we do a DB update at generate-testplans phase, the testplan yaml has the testplan-id. We can pass the entire testplan file as an input to finalize-run-testplan.

WDYT?

PR #604, opened 2018-03-21
Remove DeletionPolicy element at CF-script pre-process step

### LINKED ISSUE #423: Pre-process the Cloudformation RDS creation logic to remove "DeletionPolicy: snapshot"
**Description:**
If the deletion policy is set to snapshot, then, AWS makes a backup of the database before deletion. But, AWS has a limitation on how many manual snapshots we can create (100). 

Since testgrid re-spawns CF stacks again and again, we need to remove this policy. This may be specific to TestGrid, therefore pre-processing of the cf script is needed.

This can be found here:
https://github.com/wso2/cloudformation-is/blob/master/cloudformation-templates/pattern-1/pattern-1-with-puppet-cloudformation.template.yml#L501

Error on AWS side is this:

image

**Affected Product Version:**
0.9.0-m10

N/A

**Steps to reproduce:**
See above.

**Related Issues:**
N/A

pasindujw:
**Purpose**
If the deletion policy is set to snapshot, then, AWS makes a backup of the database before deletion. But, AWS has a limitation on how many manual snapshots we can create (100).
Since testgrid re-spawns CF stacks again and again, we need to remove this policy. 
Fixes #423 

**Approach**
Check for the DeletionPolicy element at the CF script pre-process stage and remove it.

**Security checks**

### DIFF
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/CloudFormationScriptPreprocessor.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/CloudFormationScriptPreprocessor.java
@@ -17,6 +17,8 @@
  */
 package org.wso2.testgrid.infrastructure;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.common.util.StringUtil;
 
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/CloudFormationScriptPreprocessor.java
```diff
@@ -54,15 +61,39 @@ private static String appendRandomValue(String key, String script) {
...
                         StringUtil.generateRandomString(RANDOMIZED_STR_LENGTH);
             }
             newScript.append(line).append(System.lineSeparator());
         }
         return newScript.toString();
     }
+
+    /**
+     * This method removes the elements which contains the key passed.
+     * @param key Key of the element which should be removed.
```
azinneera: New line expected

PR #619, opened 2018-03-23
Add OS as a parameter to the CF script

### LINKED ISSUE #617: OS name and version should be provided to CF scripts
**Description:**

We need to provide OS name and version as parameters to the CF scripts. There is a blocker when trying to run CF scripts for OSes other than ubuntu since the existing scripts involve logic which is depend on ubuntu (ex. apt-get). We can't switch such logic due to absence of OS and OS version in CF scripts.  

**Affected Product Version:**
0.9.0-m18

**Related Issues:**
#483

harshanL:
**Purpose**

Resolves #617 

**Security checks**

### DIFF
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
@@ -304,6 +304,13 @@ public class AWSProvider implements InfrastructureProvider {
                 cfCompatibleParameters.add(awsParameter);
             }
 
+            if (!scriptParameter.isPresent() && expected.getParameterKey().equals("OSName")) {
+                Parameter awsParameter;
+                awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
+                        .withParameterValue(getOSParameterValue(infrastructureConfig));
+                cfCompatibleParameters.add(awsParameter);
+            }
+
             scriptParameter.ifPresent(theScriptParameter -> {
                 Parameter awsParameter = new Parameter().withParameterKey(expected.getParameterKey()).
                         withParameterValue((String) theScriptParameter.getValue());
@@ -332,5 +339,9 @@ public class AWSProvider implements InfrastructureProvider {
     private String getAMIParameterValue(InfrastructureConfig infrastructureConfig)
             throws TestGridInfrastructureException {
         return amiMapper.getAMIFor(infrastructureConfig.getParameters());
+    }
+
+    private String getOSParameterValue(InfrastructureConfig infrastructureConfig) {
+        return infrastructureConfig.getParameters().getProperty("OS");
     }
 }
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
@@ -298,15 +298,13 @@ public class AWSProvider implements InfrastructureProvider {
                     .filter(input -> input.getKey().equals(expected
                             .getParameterKey())).findAny();
             if (!scriptParameter.isPresent() && expected.getParameterKey().equals("AMI")) {
-                Parameter awsParameter;
-                    awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
+                Parameter awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
                             .withParameterValue(getAMIParameterValue(infrastructureConfig));
                 cfCompatibleParameters.add(awsParameter);
             }
 
             if (!scriptParameter.isPresent() && expected.getParameterKey().equals("OSName")) {
-                Parameter awsParameter;
-                awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
+                Parameter awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
                         .withParameterValue(getOSParameterValue(infrastructureConfig));
                 cfCompatibleParameters.add(awsParameter);
             }


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -333,4 +340,8 @@ private String getAMIParameterValue(InfrastructureConfig infrastructureConfig)
             throws TestGridInfrastructureException {
         return amiMapper.getAMIFor(infrastructureConfig.getParameters());
     }
+
+    private String getOSParameterValue(InfrastructureConfig infrastructureConfig) {
+        return infrastructureConfig.getParameters().getProperty("OS");
```
kasunbg: operating_system is already a parameter. OS name and version can be properties of operating_system, yes/

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -304,6 +304,13 @@ private boolean doRelease(InfrastructureConfig infrastructureConfig, String stac
                 cfCompatibleParameters.add(awsParameter);
             }
 
+            if (!scriptParameter.isPresent() && expected.getParameterKey().equals("OSName")) {
+                Parameter awsParameter;
+                awsParameter = new Parameter().withParameterKey(expected.getParameterKey())
```
vidurananayakkara: We can write this as

```java
Parameter awsParameter = new Parameter().withParameterKey(expected.getParameterKey()).withParameterValue(getOSParameterValue(infrastructureConfig));
```

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -304,6 +304,13 @@ private boolean doRelease(InfrastructureConfig infrastructureConfig, String stac
                 cfCompatibleParameters.add(awsParameter);
             }
 
+            if (!scriptParameter.isPresent() && expected.getParameterKey().equals("OSName")) {
```
kasunbg: I think use of AMI and OSName is independent of each other. AMI is a AWS concept, while the OSName is a infrastructure_parameter property.

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
harshanL: Where do we provide operating_system as a parameter to CF?

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
harshanL: +1

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
kasunbg: It is a generic infra param that is read from the infrastructure_parameter table.

harshanL: AFAIK we do not provide operating_system as a parameter to CF since it is only used to lookup the AMI. Also current CF script does not have a parameter for OS.

kasunbg: Yes, that's the current usage. But, a CF can request for operating_system if it requires. This is just another infrastructure value..

harshanL: What do you propose?

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
harshanL: Didn't get you. Could you please elaborate more?

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
kasunbg: Like the 'database' infrastructure parameter has DBEngine and DBEngineVersion as properties, the 'operating_system' should have 'OSName' and 'OSVersion'

kasunbg: WDYT?

harshanL: +1. That should be added to the relevant db table so that it would be picked when generating test-plans. However, we need to fetch values from the infrastructure-config in the generated test-plan.

kasunbg: Cool.. I didn't understand your last sentence. :-) The infrastructure_parameter table's properties column already get processed in the generated-testplan. 

[1] https://github.com/wso2/testgrid/pull/619/files#diff-46d3c40a2de643e74df9e250c4642fa0R297

harshanL: Yes. But we only provide properties and values that are defined inside the scripts/inputParameters section to scripts. Those input-parameters are taken from the relevant infra.yaml file. So even-though the test-plan already has the OS property, it won't be used on CF. Hence we need a way to feed OS name into the CF script.

kasunbg: And, we shouldn't specialize 'OS' value; testgrid should treat it as just another parameter.

PR #620, opened 2018-03-23
Set host details from deployment.json in the proper format

### LINKED ISSUE #616: Deployment.json is not parsed properly
**Description:**
Deployment.json created by deploy.sh should be parsed in the format of tomcat host to be able to pass the correct values to replace in the jmeter scripts.

**Suggested Labels:**
Type/Bug

**Suggested Assignees:**
AsmaJ

azinneera:
**Purpose**
Contains changes related to setting the host details as `serverHost` and `serverPort` from deployment.json.
Resolves #616 

**Goals**
Provide the correct hostnames and ports for JMeter test execution.

**Approach**
Rework creation of DeploymentCreationResult in `DeploymentUtil#getDeploymentCreationResult`.

**Security checks**

### DIFF
--- a/deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
+++ b/deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
@@ -22,11 +22,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.common.Deployment;
 import org.wso2.testgrid.common.DeploymentCreationResult;
+import org.wso2.testgrid.common.Host;
+import org.wso2.testgrid.common.Port;
 import org.wso2.testgrid.common.exception.TestGridDeployerException;
+import org.wso2.testgrid.common.util.StringUtil;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This holds the utility methods used by the deployment component.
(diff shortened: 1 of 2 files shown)


deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
```diff
@@ -38,21 +43,36 @@
...
-
+        DeploymentCreationResult deploymentCreationResult = new DeploymentCreationResult();
+        List<Host> hosts = new ArrayList<>();
         ObjectMapper mapper = new ObjectMapper();
-        File file = new File(Paths.get(testPlanLocation, DeployerConstants.DEPLOYMENT_FILE).toString());
+        File file = new File(Paths.get(workspace, DeployerConstants.DEPLOYMENT_FILE).toString());
         try {
-            return mapper.readValue(file, DeploymentCreationResult.class);
+            List<Host> hostList = mapper.readValue(file, DeploymentCreationResult.class).getHosts();
+            for (Host host : hostList) {
```
vidurananayakkara: IMO we can simplify this with a lambda statement. WDYT?

harshanL: @vidurananayakkara : IMO this code won't gain anything new if we use a lambda. Is there any particular reason behind your suggestion?

deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
```diff
@@ -38,21 +43,36 @@
...
+                Host serverHost = new Host();
+                serverHost.setIp(host.getIp());
+                serverHost.setLabel("serverHost");
+                for (Port port : host.getPorts()) {
+                    Host serverPort = new Host();
+                    serverPort.setIp(String.valueOf(port.getPortNumber()));
+                    serverPort.setLabel("serverPort");
+                    hosts.add(serverPort);
+                }
+                hosts.add(serverHost);
```
harshanL: It would be better to add an explanation explaining why we have to add **serverPort** to **hosts** instead of **serverHost**.

deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
vidurananayakkara: We have 2 loops inside a try. I was thinking about the possibility of reducing that to few lines which will make the code more readable IMO as well. For the moment lets proceed with what we have.

deployment/src/main/java/org/wso2/testgrid/deployment/DeploymentUtil.java
azinneera: Added in a0f5e7e

PR #637, opened 2018-03-30
Document quick start guide for running testgrid

### LINKED ISSUE #631: Document end-to-end quickstart guide
**Description:**

We need an end-to-end guide on how a new user can use testgrid. They should be able to independently get testgrid up and running without the help of the team.

The guide at [1] details some steps on how to run testgrid. But, we need to start from where we can download testgrid dist. As I see, the steps contain, but not limited to these:

1. Download testgrid dist from $URL
2. Unzip, and copy mysql connector from $MYSQL_URL and copy to lib.
3. Populate infrastructure_parameter table with a default entry.
4. Make sure JAVA_HOME is set.

After this, users can start running their very first testgrid certification test-run.

1. Identify the repos for: infrastructure provisioning, deployment creation, and scenario testing
2. Create job-config.yaml
3. Run generate-testplan
4. Run run-testplan
5. Get reports via generate-report 

This doc can go under the docs/ folder in this repo.

[1] https://github.com/wso2/testgrid/tree/master/test-scripts/local-is-deployment

**Affected Product Version:**
0.9.0-m18

On Linux,
With JDK 8
Using MySQL DB
 
**Related Issues:**
N/A

azinneera:
**Purpose**
Adds a quick start guide to docs.
Resolves #631 

**Goals**
To provide an end-to-end quick start guide with steps for running TestGrid.

### DIFF
--- a/docs/QuickStartGuide.md
+++ b/docs/QuickStartGuide.md
@@ -1,0 +1,181 @@
+# TestGrid Quick Start Guide
+
+This provides the required steps in detail on how to run TestGrid.
+
+## Prerequisites
+
+1. Download testgrid distribution from [https://wso2.org/jenkins/job/testgrid/job/testgrid/lastSuccessfulBuild/artifact/distribution/target/WSO2-TestGrid.zip](https://wso2.org/jenkins/job/testgrid/job/testgrid/lastSuccessfulBuild/artifact/distribution/target/WSO2-TestGrid.zip)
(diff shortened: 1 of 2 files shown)


docs/QuickStartGuide.md
```diff
@@ -0,0 +1,199 @@
...
+   DB_USER_PASS
+   
+   #wum credentials
+   WUM_USERNAME
+   WUM_PASSWORD
+   
+   #AWS credentials and configurations
+   AWS_REGION_NAME
+   secretKey
+   accessKey
```
kasunbg: I think user has to create a database. If the database is named 'testgriddb' under mysql with root/root user, then do not have to configure the above DB config. 

WUM is optional? 

AWS config should be set if the user is testing AWS testing via Cloudformation.

PR #648, opened 2018-04-03
Improve reading of testgrid.yaml to skip missing properties and ignore if incomplete

### LINKED ISSUE #647: generate-test-plan command fails for incomplete testgrid.yaml
**Description:**

When testgrid.yaml does not contain scenarioConfig, the file is not ignored.
Also, TestGrid does not support config-change-sets yet. Therefore, this missing property in testgrid.yaml should be ignored which otherwise breaks the flow.

**Suggested Labels:**

Type/Bug

**Suggested Assignees:**

AsmaJ

azinneera:
**Purpose**

Contains changes related to modifying GenerateTestPlanCommand to:
  - skip missing properties in testgrid.yaml
  - ignore incomplete testgrid.yaml
Resolves #647 

**Goals**

Improve TestGrid code when reading testgrid.yaml

**Approach**

Modify `GenerateTestPlanCommand#buildTestgridYamlContent`

**Security checks**

### DIFF
--- a/core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
+++ b/core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
@@ -47,6 +47,7 @@ import org.wso2.testgrid.infrastructure.InfrastructureCombinationsProvider;
 import org.wso2.testgrid.logging.plugins.LogFilePathLookup;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.introspector.Property;
 import org.yaml.snakeyaml.nodes.NodeTuple;
 import org.yaml.snakeyaml.nodes.Tag;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -366,7 +367,7 @@ private TestgridYaml buildTestgridYamlContent(JobConfigFile jobConfigFile) {
                     TestGridConstants.TESTGRID_YAML, " is missing in ", deployRepositoryLocation));
         }
         testgridYamlContent = testgridYamlBuilder.toString().trim();
-        if (testgridYamlContent.isEmpty()) {
+        if (testgridYamlContent.isEmpty() || !testgridYamlContent.contains("scenarioConfig")) {
```
kasunbg: Seems we need scenarioConfig in lot of places. Shall we introduce a constant?

harshanL: +1

azinneera: Since this is being used only once do we need a constant?

harshanL: Then should be fine.

PR #691, opened 2018-04-18
Fix Dashboard Updating Issue

### LINKED ISSUE #661: Dashboard does not reflect the correct status of the jobs
**Description:**

Dashboard does not update properly with the database entry updates in TestGrid prod environment. We have to delete the webapp and redeploy it to see the latest results. 

**Affected Product Version:**
0.9.0-m20

**Steps to reproduce:**
1. Start a job in TestGrid prod
2. Check the dashboard for updates

lasanthaDLPDS:
**Purpose**
The purpose of this PR is to fix dashboard updating issue when database entry change. Further resolving inconsistency of redeploying web app when database entry change. 

Resolves #661 

**Goals**
Since Entity Manager retrieves data from its associated cache and if we completely bypass the entity manager cache for updating the database, the state held by the cache won't be updated to the updated database state. Hence in order to avoid this inconsistency, it is required to clear the cache when we retrieve date.

**Approach**
In order to get data forcefully from the database instead of caching, called refresh() method of Entity Manager for a returned entity. Further, if List of objects returned by a query is not itself an Entity. Therefore in that kind of case, in order to refresh a bunch of things, iterate through them and refresh them individually.

**User stories**
N/A

**Release note**

N/A

**Documentation**
N/A

**Training**
N/A

**Certification**
N/A

**Marketing**
N/A

**Automation tests**
 - Unit tests 
   N/A
 - Integration tests
   N/A

**Security checks**

**Samples**
N/A

**Related PRs**
N/A

**Migrations (if applicable)**
N/A

**Test environment**
>JDK - 1.8
> OS - UBUNTU
> DB - MySQL
> Browser Chrome
 
**Learning**
N/A

### DIFF
--- a/dao/src/main/java/org/wso2/testgrid/dao/repository/DeploymentPatternRepository.java
+++ b/dao/src/main/java/org/wso2/testgrid/dao/repository/DeploymentPatternRepository.java
@@ -22,6 +22,7 @@ import org.wso2.testgrid.common.DeploymentPattern;
 import org.wso2.testgrid.common.DeploymentPatternTestFailureStat;
 import org.wso2.testgrid.common.Status;
 import org.wso2.testgrid.common.util.StringUtil;
+import org.wso2.testgrid.dao.EntityManagerHelper;
 import org.wso2.testgrid.dao.SortOrder;
 import org.wso2.testgrid.dao.TestGridDAOException;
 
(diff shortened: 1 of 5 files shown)


dao/src/main/java/org/wso2/testgrid/dao/repository/DeploymentPatternRepository.java
```diff
@@ -130,7 +131,9 @@ public DeploymentPattern findByPrimaryKey(String id) throws TestGridDAOException
                 "dp.PRODUCT_id = '" + productId + "' GROUP BY dp.id;";
         try {
             Query query = entityManager.createNativeQuery(queryStr, DeploymentPattern.class);
-            return query.getResultList();
+            @SuppressWarnings("unchecked")
```
kasunbg: Any reason to suppress unchecked cast warning?

kasunbg:
+1 for the fix.

We are doing this knowing the performance impact. It seems this fix will incur two sql queries for each query - actual query and refresh query. 

@lasanthaDLPDS can you please create a new ticket to properly design this?

dao/src/main/java/org/wso2/testgrid/dao/repository/DeploymentPatternRepository.java
lasanthaDLPDS: Method returns List<DeploymentPattern>  and query.getResultList() return List. Therefore casting is required and can't eliminate the warning. Since this is type safe, added @SuppressWarnings("unchecked") annotation.

lasanthaDLPDS:
Created a new ticket to handle the performance impact.
https://github.com/wso2/testgrid/issues/692

PR #708, opened 2018-04-23
Improve test-run log file name

azinneera:
**Purpose**

Contains changes related to making the test-run log file name more context aware.

**Goals**

Improve logging implementation.

**Approach**

The new log file will be named as `<testgrid-home>/<product-name>/logs/<deployment_pattern>_<infrastructure-combination/uuid>_<build-number>.log`

**Security checks**

### DIFF
--- a/web/src/main/java/org/wso2/testgrid/web/api/LogEventSocketMediator.java
+++ b/web/src/main/java/org/wso2/testgrid/web/api/LogEventSocketMediator.java
@@ -26,11 +26,13 @@ import org.wso2.testgrid.common.util.StringUtil;
 import org.wso2.testgrid.common.util.TestGridUtil;
 import org.wso2.testgrid.dao.TestGridDAOException;
 import org.wso2.testgrid.dao.uow.TestPlanUOW;
+import org.wso2.testgrid.logging.plugins.LogFilePathLookup;
 import org.wso2.testgrid.web.utils.FileWatcher;
 import org.wso2.testgrid.web.utils.FileWatcherException;
 
 import java.io.IOException;
 import java.nio.file.Path;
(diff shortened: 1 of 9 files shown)


web/src/main/java/org/wso2/testgrid/web/api/LogEventSocketMediator.java
```diff
@@ -130,7 +132,13 @@ private void runLogTailer(TestPlan testPlan, Session session)
      * @throws TestGridException thrown when error on calculating the log file path
      */
     private Path getLogFilePath(TestPlan testPlan) throws TestGridException {
-        return TestGridUtil.getTestRunWorkspace(testPlan, false).resolve(TestGridConstants.TEST_LOG_FILE_NAME);
+        //Create logging directory
+        Path testRunDirectory = TestGridUtil.getTestRunWorkspace(testPlan);
+        String[] subStrings = testRunDirectory.toString().split(TestGridConstants.FILE_SEPARATOR, 2);
+        String logFileName = subStrings[1].replaceAll(TestGridConstants.FILE_SEPARATOR, "_")
```
kasunbg: This could lead to ArrayIndexOutofBoundsException I think. We can't always guarantee that the array size is two. 

I think what you really require is the testgrid job name. I think we should pass the name from jenkinsfile into the testgrid from the job-config.yaml. wdyt?

azinneera: We will require to derive the name in all commands. Therefore, if we use the job-config.yaml, we will have to either read it in every command or persist the job name at generate-test-plan command execution. We can make use of the TestGridUtil#deriveLogFilePath for this. WDYT?

PR #729, opened 2018-05-02
Externalize JMeter from TestGrid

### LINKED ISSUE #701: Externalize JMeter engine from the TestGrid
**Description:**

JMeter engine has to be externalized to achieve higher test success rate and fewer dependency problems when running scenario tests. For this we also need to design a proper output format from scenario results in which we can process and generate a report.

harshanL:
**Purpose**

Resolves #701 

**Approach**

JMeter has to be installed locally and JMETER_HOME env var should be added to the environment. TestGrid will invoke a scenario.sh script inside each scenario which in-turns use installed JMeter and run jmx files. 

**Security checks**


automation/src/main/java/org/wso2/testgrid/automation/reader/JMeterTestReader.java
```diff
@@ -25,54 +25,55 @@
...
  * Reader for reading Jmeter test files of the standard maven structure.
  */
 public class JMeterTestReader implements TestReader {
 
-    private static final String JMTER_SUFFIX = ".jmx";
     private static final String JMETER_TEST_PATH = "jmeter";
     private static final String SHELL_SUFFIX = ".sh";
     private static final String PRE_STRING = "pre-scenario-steps";
     private static final String POST_STRING = "post-scenario-steps";
+    private static final String SCENARIO_SCRIPT = "scenario.sh";
```
azinneera: Not required in this PR but shall we change this to read the pre, post script names and scenario script names to read from testgrid.yaml just like how we read deploy script and infra script?

harshanL: But the issue is there will be only one deploy script & infra script per test-plan and there will be multiple scenario scripts. So wouldn't it be hard to add the scenario script name in each test scenario rather than going to a convention?

automation/src/main/java/org/wso2/testgrid/automation/parser/FunctionalTestResultParser.java
```diff
@@ -0,0 +1,154 @@
...
+            }
+        } catch (XMLStreamException e) {
+            throw new JMeterResultParserException("Unable to parse the scenario-results file of TestScenario :" +
+                    testScenarioName, e);
+        } catch (FileNotFoundException e) {
+            throw new JMeterResultParserException("Unable to locate the scenario-results file.", e);
+        } finally {
+            try {
+                if (inputStream != null) {
+                    inputStream.close();
```
kasunbg: Shall we use try-with-resources to reduce boiler-plate code?

automation/src/main/java/org/wso2/testgrid/automation/parser/FunctionalTestResultParser.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+            } catch (IOException e) {
+                throw new JMeterResultParserException("Unable to close the input stream of scenario results file of " +
+                        "the TestScenario : " + testScenarioName, e);
+            }
+        }
+    }
+
+    private TestCase addAttributes(StartElement startElement) {
```
kasunbg: Missing javadocs.
How about we change the method name from addAttributes to something more meaningful? Say, buildTestCaseFrom? The input parameter name can be something like jmeterSampleElement. wdyt?

automation/src/main/java/org/wso2/testgrid/automation/parser/JMeterResultParserFactory.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (XMLStreamException e) {
+            throw new JMeterResultParserException("Unable to parse the scenario-results file of the test scenario : " +
+                    "'" + testScenarioName + "'", e);
+        } catch (FileNotFoundException e) {
+            throw new JMeterResultParserException("Unable to locate the scenario-results file of the test scenario : " +
+                    "'" + testScenarioName + "'", e);
+        } finally {
+            if (inputStream != null) {
+                try {
+                    inputStream.close();
```
kasunbg: try-with-resources?

automation/src/main/java/org/wso2/testgrid/automation/parser/JMeterResultParserFactory.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * This method returns an instance of {@link JMeterResultParser} to parse the given JMeter result file.
+     *
+     * @param testScenario {@link TestScenario}  TestScenario object associated with the tests
+     * @param testLocation {@link String}        Location of the scenario test directory
+     * @return a JMeterResultParser {@link JMeterResultParser} to parse the given JTL file
+     * @throws JMeterResultParserException {@link JMeterResultParserException} when an error occurs while obtaining the
+     * instance of the parser
+     */
+    public static Optional<JMeterResultParser> getParser(TestScenario testScenario, String testLocation) throws
```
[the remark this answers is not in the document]
harshanL: This is not parsing the entire content of jtl. But this is required to obtain the correct parser for parsing the JTL file. This is merely based on the start element of the JTL file.

automation/src/test/resources/test-grid-is-resources/SolutionPattern22/Tests/jmeter/empty.jmx
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<jmeterTestPlan version="1.2" properties="3.2" jmeter="3.3 r1808647">
```
kasunbg: I think we should still have jmeter based unit tests within testgrid after externalizing the jmeter as well.
Test phase is a core part of testgrid. We need some kind of tool to verify that phase. Jmeter is the ideal candidate for that. 

Our previous impl downloaded jmeter dependency via maven. In the new impl, we are invoking jmeter via script. So, for that, we have to download extract it manually. This can be done via the TestNG @BeforeSuite step.

Shall we do it here or create a separate ticket to track this?

automation/src/main/java/org/wso2/testgrid/automation/parser/FunctionalTestResultParser.java
```diff
@@ -0,0 +1,154 @@
...
+        TestCase testCase = new TestCase();
+        testCase.setTestScenario(this.testScenario);
+        Iterator<Attribute> attributes = startElement.getAttributes();
+
+        while (attributes.hasNext()) {
+            Attribute attribute = attributes.next();
+            if (TEST_NAME_ATTRIBUTE.equals(attribute.getName().getLocalPart())) {
+                testCase.setName(attribute.getValue());
+            } else if (TEST_STATUS_ATTRIBUTE.equals(attribute.getName().getLocalPart())) {
```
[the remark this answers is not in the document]
harshanL: We could set the TestCase as successful if the "assertionResult" node does not contain any error or failure message. But the attribute "ec" was there in every test case in identity-integration tests.

automation/src/test/resources/test-grid-is-resources/SolutionPattern22/Tests/scenario-results.jtl
```diff
@@ -0,0 +1,3988 @@
...
+                document.write(&apos;&lt;/td&gt;&lt;/tr&gt;&lt;/div&gt;&apos;);
+            }
+            document.write(&apos;&lt;/tbody&gt;&apos;);
+        }
+
+    &lt;/script&gt;
+
+&lt;/head&gt;
+&lt;!-- ===================================== END HEADER ===================================== --&gt;
+&lt;body&gt;&lt;a id=&quot;top-of-page&
```
kasunbg: The JTL is 4000 lines it seems.. :) I don't think we need one this big for unit tests.

automation/src/test/resources/test-grid-is-resources/SolutionPattern22/Tests/jmeter/empty.jmx
harshanL: +1. New issue [1] created for it.

[1]. https://github.com/wso2/testgrid/issues/733

automation/src/test/resources/test-grid-is-resources/SolutionPattern22/Tests/scenario-results.jtl
harshanL: New issue [1] created.

[1]. https://github.com/wso2/testgrid/issues/734

automation/src/main/java/org/wso2/testgrid/automation/reader/JMeterTestReader.java
kasunbg: I guess we can use this name as the default. We can provide an option to over-ride this via testgrid.yaml if required. We should not dictate the script names of test writers. Our only requirement should be the .testgrid.yaml. :) 

WDYT?

harshanL: +1

automation/src/main/java/org/wso2/testgrid/automation/parser/FunctionalTestResultParser.java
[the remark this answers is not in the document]
kasunbg: OK, if you think assertionResult node is a better estimate, then we should use that.
May be the 'ec' attribute was there for identity jmeter scripts because of some convention they follow. It's not fool-proof as I've seen.

Issue #722, opened 2018-04-27
Implement initial PoC to execute custom commands inside deployment's VMs


kasunbg:
**Description:**
We need to have the initial PoC for $subject ready to run in testgrid. The initial idea is to execute the custom commands needed by scenario test writers. For example, a core requirement when doing long-running tests is to intermittently slow-down the network. To do that, the script writers will need to execute OS commands in those VMs.

This initial PoC is targeted for AWS deployments. This is implemented as part of the effort for getting long running tests to work in Testgrid #688. Initially, this'll be targeted against DAS #572.

**Affected Product Version:**
0.9.0-m22

**OS, DB, other environment details and versions:**    
AWS

**Steps to reproduce:**
N/A

**Related Issues:**
N/A

Issue #688, opened 2018-04-16
Support Long running tests in testgrid


kasunbg:
**Description:**
Currently, we support functional testing in testgrid. We need to support long running tests as well..

Long running tests are a different beast compared to functional testing. Here, the focus is not on verifying the functionality, but on non-functional requirements such as:

# Server's CPU, Memory utilization are healthy over a period of time.
# Server functions as expected over the entire period of time tested. 

Further, a small percentage of errors can be disregarded if it is under an accepted threshold. 
For example, in a deployment where we only have two-node minimal HA cluster, there can of course be functionality failures/errors if the peak traffic is too high. What's really important is whether the server recovers after this burst traffic.

Research on designing the long running tests process will be carried out by the STL team. Testgrid need to closely work with STL team to make sure the correct runtime support is provided from testgrid runtime.

CC @chaminda @dilinisg

**Affected Product Version:**
0.9.0-m22

**OS, DB, other environment details and versions:**    
IS/APIM/EI
AWS

**Steps to reproduce:**
N/A

**Related Issues:**
N/A

Issue #572, opened 2018-03-14
Integrate DAS into Testgrid


kasunbg:
**Description:**
DAS team has sparked their interest to add their product into testgrid. 

I do not think they have basic installation scripts yet. Therefore, I think we should present them with some base cloudformation scripts to deploy an one node DAS instance in AWS. 

We should coordinate with solution test labs team and get a basic set of JMeter scenario test scripts as well for the initial onboarding process.

**Affected Product Version:**
0.9.0-m16

**OS, DB, other environment details and versions:**    
DAS 3.1.0
AWS

**Steps to reproduce:**
N/A

**Related Issues:**
#416 #417 #419

PR #747, opened 2018-05-09
Implementation of initial PoC to execute custom commands inside deployment's VMs

charithag:
**Purpose**

This PoC is done to execute custom commands needed by scenario test writers though an agent deployed in the target VM instance. For example, a core requirement when doing long-running tests is to intermittently slow-down the network. To do that, the script writers will need to execute OS commands in those VMs.

This initial PoC is targeted for AWS deployments. This is implemented as part of the effort for getting long running tests to work in Testgrid #688. Initially, this'll be targeted against DAS #572.

Git issue link: https://github.com/wso2/testgrid/issues/722

**Goals**
1. Apply config-change-sets of scenario test writers into the product instances.
2. Execute arbitrary commands in these remote VMs as required by scenario tests. For example, long running tests may need to connect to the deployment to simulate network slowness (via tc)
3. Stop/Restart WSO2 product instances.

In order to gain these capabilities, we can deploy an Agent on VM and control the VM environment.

**Approach**
This is the high-level overview of the design:
testgrid vm agent design

Agent application developed as a Java SE stand alone app which runs as a Linux service with support of shell script and included to AMIs. Agent configured to communicate with remote session web app using web sockets. Remote management web application deployed on a tomcat 8.x will manage the agents and expose REST API interface to list online agents and send commands to execute.

**User stories**
The scenario test scripts need to connect to the deployment to mutate it as necessary. One common usage is simulating real production environment for the deployment to certify long running behavior of WSO2 servers.

For this purpose, Testgrid can expose a REST API and once operation received to REST API, it will immediately communicated to the agent and results coming from agent will send to CLI tool as the response. The scenario test writers can use this CLI tool to talk to the product deployment via testgrid.

**Release note**
Implementation of initial PoC to execute custom commands inside deployment's VMs.

**Documentation**
Agent should be initialized before running by using init.sh script when executing the user data section in the CF script as follows.

> ./init.sh \<remote management endpoint\> \<aws region\> \<test plan id\> \<cloud provider\>

Above initialization command will be added to CF script via AWSProvider class in the test grid core. In order to add above command, CF script should contains `CustomUserData` parameter as per https://github.com/harshanL/cloudformation-is/pull/2/files.

Also in order to provide remote management endpoint for agents, it is needed to define it in the config.properties file in the test grid home.
Ex:
> #Remote Management Server endpoint
> REMOTE_MANAGEMENT_EP=ws://ec2-34-232-211-33.compute-1.amazonaws.com:8080/remote-session/v0.9/agent/

Here is the REST API endpoint provided via remote session web app.

1. `GET api/management/agents ` - List all active agents
> curl -X GET http://ec2-34-232-211-33.compute-1.amazonaws.com:8080/remote-session/v0.9/api/management/agents
2. `GET api/management/test-plan/{test plan id}/agents` - List all active agents for test plan
> curl -X GET http://ec2-34-232-211-33.compute-1.amazonaws.com:8080/remote-session/v0.9/api/management/test-plan/6f491ddd-1324-49ce-aeaa-694e52512d21/agents
3. `POST api/management/test-plan/6f491ddd-1324-49ce-aeaa-694e52512d21/agent/wso2-is-node/operation` - Send shell command operation to agent
> curl -X POST \
>  http://ec2-34-232-211-33.compute-1.amazonaws.com:8080/remote-session/v0.9/api/management/test-plan/{test plan id}/agent/{agent name}/operation \
>  -H 'content-type: application/json' \
>  -d '{"code":"SHELL", "request":"service testgrid-agent status"}'

**Security checks**

**Test environment**
Ubuntu 16.04, JDK 1.8, MySQL 5.7, Apache Tomcat 8.5.23


kasunbg:
@charithag +1 for the design. I think all the needed bits are described in the diagram and the guideline. Please find some comments.

1. Shall we protect the new API with some authentication mechanism? Since the requests go over the internet, we need to secure them. 
2. Any specific reason to name the webapp as "remote session"? 
3. As I understand, this API is not about managing the agents. Rather, it does the actual operations against agents. As I see, management is about Agent registration/de-registration and any other operations where we actually manage the agents. 
At the same time, we need to clearly separate out the API intended for testgrid's clients (for ex. the scenario tests) and the API intended for agents.
So, shall we re-think the API a bit more?
4. Could you also list the agent registration API?
5. Regarding the below command:
`./init.sh <remote management endpoint> <aws region> <test plan id> <cloud provider>`
I did not understand who provide the `init.sh` to agents. :) Is it actually a part of testgrid-cli?
And, the `aws-region` is specifically about AWS. So, there's no need to take cloud-provider as another argument? We'll need to generalize the parameters if we are writing this script to be cloud-provider neutral imo.

WDYT?

agent/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
+    </dependencies>
+</project>
```
kasunbg: Missing new line at EOF.

agent/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <configuration>
+                    <appendAssemblyId>false</appendAssemblyId>
+                    <finalName>agent</finalName>
```
kasunbg: testgrid-agent?

agent/src/main/bin/init.sh
```diff
@@ -0,0 +1,25 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+echo "wsEndpoint=$1" > /opt/testgrid/agent-config.properties
+echo "region=$2" >> /opt/testgrid/agent-config.properties
+echo "testPlanId=$3" >> /opt/testgrid/agent-config.properties
+echo "provider=$4" >> /opt/testgrid/agent-config.properties
```
kasunbg: Ideally, we should parse the input arguments via [getopts](https://www.mkssoftware.com/docs/man1/getopts.1.asp). Shall we add a TODO item within the file itself and/or raise a ticket, so we know about the limitation? :)

agent/src/main/bin/testgrid-agent
```diff
@@ -0,0 +1,170 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+# chkconfig: 345 99 05
+# description: TestGrid Agent daemon script
+#
+# A non-SUSE Linux start/stop script for Java daemons.
+#
+# Set this to Java installation
+JAVA_HOME=/usr/bin/
```
kasunbg: It seems weird why the JAVA_HOME would be at /usr/bin. :)

agent/src/main/bin/testgrid-agent
```diff
@@ -0,0 +1,170 @@
...
+
+scriptFile=$(readlink -fn $(type -p $0))             # the absolute, different path of this script file
+scriptDir=$(dirname $scriptFile)                     # absolute path of the script directory
+serviceNameLo="testgridAgent"                        # service name with the first letter in lowercase
```
kasunbg: As I've seen, dash is the standard separator in unix world.

agent/src/main/bin/testgrid-agent
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
kasunbg: The purpose of this script is a bit unclear.

agent/src/main/bin/testgrid-agent
```diff
@@ -0,0 +1,170 @@
...
+serviceName="TestGrid Agent"                         # service name
+serviceUser="root"                                   # OS user name for the service
+serviceGroup="root"                                  # OS group name for the service
+applDir="/opt/testgrid/agent"                        # home directory of the service application
+serviceUserHome="/home/$serviceUser"                 # home directory of the service user
+serviceLogFile="$applDir/$serviceNameLo.log"         # log file for StdOut/StdErr
+maxShutdownTime=15                                   # maximum number of seconds to wait for the daemon to terminate normally
+pidFile="$applDir/$serviceNameLo.pid"                # name of PID file (PID = process ID number)
+javaCommand="java"                                   # name of the Java launcher without the path
+javaArgs="-cp agent.jar:libs/* org.wso2.testgrid.agent.AgentApplication"  # arguments for Java launcher
```
kasunbg: What were the assumptions to set the classpath as "agent.jar:libs/*"? For example, do you assume a testgrid dist should be there in this location?

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+
+package org.wso2.testgrid.agent;
+
+import com.google.gson.Gson;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wso2.testgrid.agent.beans.OperationRequest;
+import org.wso2.testgrid.agent.beans.OperationResponse;
+import org.wso2.testgrid.agent.listners.OperationResponseListener;
```
kasunbg: Typo in listners package name.

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Agent bootstrap class.
+ *
+ * @since 1.0.0
+ */
+public class AgentApplication implements OperationResponseListener {
+
+    private static final Logger logger = LoggerFactory.getLogger(AgentApplication.class);
+    private static final String AGENT_PROP_FILE = File.separator + "opt" + File.separator + "testgrid" +
+            File.separator + "agent-config.properties";
```
kasunbg: We'll need to document about this file?

Let's try to use java.nio.Paths next time! :)

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
```diff
@@ -0,0 +1,132 @@
...
+                } catch (IOException e) {
+                    logger.error("Error occurred when closing prop file stream: " + e.getMessage(), e);
+                }
+            }
+        }
+
+        try {
+            logger.info("Agent ID: " + agentId);
+            // open web socket
+            clientEndPoint = new ClientEndpoint(new URI(wsEndpoint + agentId));
```
kasunbg: This assumes the wsEndpoint will end with a '/'.

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
```diff
@@ -0,0 +1,132 @@
...
+                public void handleMessage(String message) {
+                    logger.info("Operation received: " + message);
+                    OperationRequest operationRequest = new Gson().fromJson(message, OperationRequest.class);
+                    OperationExecutor operationExecutor = new OperationExecutor(AgentApplication.this);
+                    operationExecutor.executeOperation(operationRequest);
+                }
+
+                @Override
+                public void handleClose(CloseReason reason) {
+                    logger.warn("Client disconnected. Retrying to connect.");
```
kasunbg: Shall we print the CloseReason?

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
```diff
@@ -0,0 +1,132 @@
...
+        try {
+            if (new File(AGENT_PROP_FILE).exists()) {
+                input = new FileInputStream(AGENT_PROP_FILE);
+                prop.load(input);
+                wsEndpoint = prop.getProperty("wsEndpoint");
+                //aws:us-west-1:6718f976-7150-316d-9ec6-db29b7a2675f:i-0658922f45d25856f
+                agentId = prop.getProperty("provider") + ":" + prop.getProperty("region") + ":" +
+                        prop.getProperty("testPlanId") + ":" + prop.getProperty("instanceId");
+            } else {
+                logger.warn("Agent configurations not found.");
```
kasunbg: The name and where you tried to find it out is missing.

agent/src/main/java/org/wso2/testgrid/agent/OperationExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    private String executeCommand(String command) {
+        StringBuilder output = new StringBuilder();
+        Process p;
+        BufferedReader reader = null;
+        try {
+            p = Runtime.getRuntime().exec(command);
+            p.waitFor();
+            reader = new BufferedReader(new InputStreamReader(p.getInputStream(), UTF_8));
```
kasunbg: The getInputStream returns the stdout. We need stderr as well. 

Please see a sample impl on how to run ProcessBuilder - https://github.com/wso2/testgrid/blob/291abadfe83d405a428c2a674d2b2fa6dc98a5ae/common/src/main/java/org/wso2/testgrid/common/ShellExecutor.java#L102

agent/src/main/bin/init.sh
charithag: ATM, init.sh call inject into User Data section of the CF by TG. So will add TODO here.

agent/pom.xml
charithag: Agent is located in /opt/testgrid/ directory and having 'testgrid-' will add redundancy in the absolute path

agent/src/main/bin/testgrid-agent
charithag: This is using to facilitate agent as a int.d compatible Linux service

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: This will be a automatically generated file by init.sh script which is not required user awareness.

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: It is already printed with onClose call back in ClientEndpoint.

agent/src/main/bin/testgrid-agent
kasunbg: Going forward I think we should provider a single uber jar that we can just download and run. This allows easier setting up of slaves.

agent/src/main/assembly/bin.xml
```diff
@@ -0,0 +1,50 @@
...
+        <file>
+            <source>src/main/bin/testgrid-agent</source>
+            <fileMode>0755</fileMode>
+        </file>
+        <file>
+            <source>src/main/bin/init.sh</source>
+            <fileMode>0755</fileMode>
+        </file>
+    </files>
+</assembly>
```
harshanL: Missing a new line.

agent/pom.xml
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/bin/testgrid-agent
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/bin/testgrid-agent
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: Used java.nio.paths with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: Handled endpoint with or without '/' with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/AgentApplication.java
charithag: Added conf file absolute path in to the log with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/OperationExecutor.java
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/OperationExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+                output.append(line).append("\n");
+            }
+        } catch (InterruptedException | IOException e) {
+            String message = "Error occurred when executing command: " + command;
+            logger.error(message, e);
+            return message;
+        } finally {
+            if (reader != null) {
+                try {
+                    reader.close();
```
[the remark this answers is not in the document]
charithag: Used try-with-resources in commit https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad

agent/src/main/java/org/wso2/testgrid/agent/websocket/ClientEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+    private boolean hasClientConnected = false;
+    private boolean isShuttingDown = false;
+
+    public ClientEndpoint(URI endpointURI) {
+        this.endpointURI = endpointURI;
+    }
+
+    public void connectClient() {
+        try {
+            ClientManager client = ClientManager.createClient();
```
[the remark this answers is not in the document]
charithag: Yes

agent/src/main/assembly/bin.xml
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/af9c647d9f201e6d77d5044846ddeab7b6c391c1

common/src/main/java/org/wso2/testgrid/common/TestGridConstants.java
```diff
@@ -47,5 +47,6 @@
 
     public static final String WUM_USERNAME_PROPERTY = "WUMUsername";
     public static final String WUM_PASSWORD_PROPERTY = "WUMPassword";
+    public static final String CUSTOM_USER_DATA = "CustomUserData";
```
kasunbg: Testgrid core is independent of the cloud providers. Seems this shud be added to AWSProvider.

charithag:
@kasunbg,

Please find my inline responses for you comments.

1. Shall we protect the new API with some authentication mechanism? Since the requests go over the internet, we need to secure them.
> Added basic authentication with commit https://github.com/wso2/testgrid/pull/747/commits/c5d3bd66a54af2695574fccd6b7cd713919393db
2. Any specific reason to name the webapp as "remote session"?
> Renamed web app name to 'Deployment tinkerer'
3. As I understand, this API is not about managing the agents. Rather, it does the actual operations against agents. As I see, management is about Agent registration/de-registration and any other operations where we actually manage the agents.
> Updated API path params in the REST APIs with commit https://github.com/wso2/testgrid/pull/747/commits/b2812b0818ed38457361934f505c931e587efcad
4. At the same time, we need to clearly separate out the API intended for testgrid's clients (for ex. the scenario tests) and the API intended for agents.
> Already done.
5. So, shall we re-think the API a bit more?
> Updated API endpoints.
6. Could you also list the agent registration API?
> Updated the PR description.
7. Regarding the below command:
./init.sh <remote management endpoint> <aws region> <test plan id> <cloud provider>
I did not understand who provide the init.sh to agents. :) Is it actually a part of testgrid-cli?
> init.sh will execute from user data section in the CF. Execution command will be added in to CF by AWSProvider in the TG core.
8. And, the aws-region is specifically about AWS. So, there's no need to take cloud-provider as another argument? We'll need to generalize the parameters if we are writing this script to be cloud-provider neutral imo.
> In agent-config.properties this will store as 'region'.

common/src/main/java/org/wso2/testgrid/common/TestGridConstants.java
charithag: Fixed with commit https://github.com/wso2/testgrid/pull/747/commits/8dffa11a3baf2ab1b238094e4301c57e85a04be7

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/beans/OperationResponse.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.wso2.testgrid.deployment.tinkerer.beans;
+
+/**
+ * This class holds operation response data.
+ *
+ * @since 1.0.0
+ */
+public class OperationResponse extends Operation {
```
kasunbg: I would expect that a sub-class of Operation does a specialized operation. For example, something like ExecuteShellOperation, LogStreamOperation, FileArchiveOperation etc. 

The purpose of OperationResponse is a bit unclear. :)

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
```diff
@@ -0,0 +1,82 @@
...
+import org.wso2.testgrid.common.util.TestGridUtil;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Implementation of the AWS cloud provider functionality.
```
kasunbg: We already have another AWSProvider in testgrid infrastructure module. IMO, this javadoc does not distinguish this class's funtionality. wdyt?

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    static {
+        try {
+            configFilePath = TestGridUtil.getConfigFilePath();
+        } catch (IOException e) {
+            logger.error("Error occurred while getting the config path.", e);
+        }
+    }
+
+    @Override
+    public String getInstanceName(String region, String instanceId) {
```
kasunbg: *getEC2InstanceName.

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        DescribeInstancesRequest request = new DescribeInstancesRequest();
+        request.setInstanceIds(instanceIds);
+
+        DescribeInstancesResult result = amazonEC2.describeInstances(request);
+        Optional<Tag> tagOptional = result.getReservations()
+                .stream()
+                .flatMap(reservation -> reservation.getInstances().stream())
+                .flatMap(instance -> instance.getTags().stream())
+                .filter(tag -> "Name".equals(tag.getKey()))
+                .findFirst();
```
kasunbg: Can we combine the last #filter and #findFirst via #anyMatch terminal operation? anyMatch is equivalent to filter+findAny combination.

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/HttpSessionConfigurator.java
```diff
@@ -0,0 +1,56 @@
...
+     * <p>If the developer does not override this method, no further
+     * modification of the request and response are made by the implementation.
+     *
+     * @param config the configuration object involved in the handshake
+     * @param request  the opening handshake request.
+     * @param response the proposed opening handshake response
+     */
+    public void modifyHandshake(ServerEndpointConfig config, HandshakeRequest request, HandshakeResponse response) {
+        Map<String, List<String>> httpHeaders = request.getHeaders();
+        config.getUserProperties().put(Constants.HTTP_HEADERS, httpHeaders);
```
kasunbg: Does a config object created per HTTP request?

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/SessionManager.java
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * This class manage sessions of agents corresponding to agent ids.
+ *
+ * @since 1.0.0
+ */
+public class SessionManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(SessionManager.class);
+
+    private static SessionManager sessionManager = new SessionManager();
```
kasunbg: final?

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
charithag: This method should be generic as it defined in the Provider interface and intended to be generic across multiple providers like Azure, GCE

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/SessionManager.java
```diff
@@ -0,0 +1,210 @@
...
+    public Agent getAgent(String testPlanId, String instanceName) {
+        Optional<Map.Entry<String, Agent>> agentOptional = agents.entrySet().stream()
+                .filter(entry -> {
+                    Agent agent = entry.getValue();
+                    if (agent != null && testPlanId != null && instanceName != null) {
+                        return testPlanId.equals(agent.getTestPlanId()) && instanceName.equals(agent.getInstanceName());
+                    }
+                    return false;
+                }).findFirst();
+        return agentOptional.map(Map.Entry::getValue).orElse(null);
```
kasunbg: Why not return Optional directly?

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/HttpSessionConfigurator.java
charithag: yes

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/SessionManager.java
charithag: In getAgent method, we are intended to return an Agent object. The optional contains Map.Entry<String, Agent> object.

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
charithag: Stream#anyMatch() returns a boolean while Stream#findFirst() returns an object which matches the predicate. Here we wan't to have the Optional with the Tag object, not the evaluation of tag existence.

https://stackoverflow.com/a/44179797/1391687

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/beans/OperationResponse.java
charithag: Operation request contains the request coming from the user while Operation response contains the response for operations. Both are forms of abstract operation and shares same data and functionalities while operation response holds the response for the operation and operation request holds the original request from the user.
In here we don't need to have different sub classes to model different operations as long as operation it self supported to cater different types of operations with operation code attribute.

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/SessionManager.java
charithag: Added with commit https://github.com/wso2/testgrid/pull/747/commits/79a96b3a95c101249b13c1bab72b0e1faf640a2a

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/api/DeploymentTinkerer.java
```diff
@@ -0,0 +1,153 @@
...
+
+package org.wso2.testgrid.deployment.tinkerer.api;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.wso2.testgrid.deployment.tinkerer.beans.Agent;
+import org.wso2.testgrid.deployment.tinkerer.beans.ErrorResponse;
+import org.wso2.testgrid.deployment.tinkerer.beans.OperationRequest;
+import org.wso2.testgrid.deployment.tinkerer.utils.Constants;
+import org.wso2.testgrid.deployment.tinkerer.utils.SessionManager;
```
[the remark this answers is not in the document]
charithag: Moved with commit https://github.com/wso2/testgrid/pull/747/commits/79a96b3a95c101249b13c1bab72b0e1faf640a2a

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/HttpSessionConfigurator.java
kasunbg: OK, I didn't get it when reading this - https://docs.oracle.com/javaee/7/api/javax/websocket/server/ServerEndpointConfig.html

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/utils/SessionManager.java
kasunbg: Yes, so how about Optional<Agent>? Optional is recommended to be used for return types.

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/providers/AWSProvider.java
charithag: Modified description with commit https://github.com/wso2/testgrid/pull/747/commits/ec6f546ca8f847d53674f0782bcf7cd276bbb1f5

deployment-tinkerer/src/main/java/org/wso2/testgrid/deployment/tinkerer/beans/OperationResponse.java
kasunbg: This is an interesting discussion. Let's continue this outside this PR. :)

kasunbg: I think the Operations need to be given first-class status. It should not hidden inside somewhere as an enum. An operation need to be a concrete class with an #execute method that knows how to execute that operation. I think execute-shell, ping, log-archive are all operations that need to have their on #execute implementations.

At the same time, the request/response analogy that you bring up is quite same as [HttpServletRequest](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html) and [HttpServletResponse](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html). I do not see the parent-child hierarchy you mention in there.

Going further, shall we think how we improve this? I think we can model this similar to how Axis2 cluster messages work. In Axis2, Cluster messages is a way to send instructions and some state to other nodes. So, how about a class-hierarchy like this that I based on your impl and axis2 cluster message design:

```java
interface Operatable {
    String execute();
}

class OperationResponse {
  String uuid;
  public String getUUID() {}
}

class ExecuteShellOperation implements Operatable {
  public String execute() {}
}
class PINGOperation implements Operatable {
  public String execute() {}
}

class ExecuteShellOperationResponse extends OperationResponse {}
class PINGOperationResponse extends OperationResponse {}

```

Then, you can send a message to an agent like follows: 
```java
class HeartBeatChecker {
  public void sendHeartBeat(SessionManager sessionManager, String agentId) {
    Operatable op = new PINGOperation();
    String uuid = sessionManager.sendOperation(op);
    OperationResponse opResp = sessionManager.getResponse(uuid); // blocking as in j.u.c.Future.get(). We can provide an async method as well.
  }
}
```

We can obviously fine-tune this. WDYT about the overall design?

core/pom.xml
```diff
@@ -114,5 +114,10 @@
             <artifactId>org.jacoco.agent</artifactId>
             <classifier>runtime</classifier>
         </dependency>
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>6.0.6</version>
```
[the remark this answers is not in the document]
charithag: Fixed with https://github.com/wso2/testgrid/pull/747/commits/3ee4e2250fca65e3fe7c4a4abab9a99cd8d16a72

PR #761, opened 2018-05-21
Generate infra combination only for product job supporting infra

### LINKED ISSUE #645: Infrastructure combinations should only generated for infrastructures product job support
**Description:**
We now have multiple infrastructure types and infrastructure items in our catalog. We run all our jobs against the entire catalog. Then, the dashboard shows build status against infrastructure combinations that are made up of all these infrastructure items. But the truth is that not all our jobs are ready for this. Only IS jobs supports the entirety of the infrastructure items. So, it's a bit misleading.

I think we need to bind each product to infrastructure items it can support (at testgrid level). 

**Affected Product Version:**
0.9.0-m19
APIM testgrid job

**Steps to reproduce:**
1. Run APIM testgrid job.
2. You'll see that it was run for number of infra combinations in the dashboard.
3. But underneath, the testgrid input parameters were probably skipped.

**Suggested Fix**
I think the supported infrastructure items should be listed in the testgrid.yaml of infrastructure repositories of the product. 

**Related Issues:**
N/A

lasanthaDLPDS:
**Purpose**
The purpose of this PR is to resolve #645

**Goals**
In order to achieve this task modified **testgrid.yaml** file and added new attributes called exclude and include. If a user has defined exclude attribute, before generating infra combinations, filter the infra set and removed excluding set from the infra set. If a user has defined include attribute, before generating infra combinations, filter the infra set and only include defined infra, if and only if that infra is in the **infrastructure_parameter** table. If both exclude and include options are defined in the **testgrid.yaml** file only considers excluding option.

**Approach**
tgpr1

**User stories**

N/A

**Release note**

N/A

**Documentation**
N/A

**Training**
N/A

**Certification**
N/A

**Marketing**
N/A

**Automation tests**
 - Unit tests 
   N/A
 - Integration tests
   N/A

**Security checks**

**Samples**
N/A

**Related PRs**
N/A

**Migrations (if applicable)**
N/A

**Test environment**
>JDK - 1.8
>OS - UBUNTU 18.04
>DB - Mysql
 
**Learning**
N/A


core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -131,4 +136,44 @@ private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombinatio
...
+    private Set<InfrastructureValueSet> filterInfrastructures(Set<InfrastructureValueSet> infrastructures,
+            TestgridYaml testgridYaml) {
+
+        List<String> excludes = testgridYaml.getInfrastructureConfig().getExcludes();
+        List<String> includes = testgridYaml.getInfrastructureConfig().getIncludes();
+        Set<InfrastructureValueSet> selectedInfraValSet = ConcurrentHashMap.newKeySet();
+
+        if (excludes != null && !excludes.isEmpty()) {
+            infrastructures.forEach(infrastructureValueSet -> {
+                Set<InfrastructureParameter> selectedSet = infrastructureValueSet.getValues().stream()
```
hasuniea: Does infrastructureValueSet.getValues() method always return a value, unless there will be a possibility to happen NPE

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -131,4 +136,44 @@ private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombinatio
...
+                        .filter(infrastructureParameter -> !excludes.contains(infrastructureParameter.getName()))
+                        .collect(Collectors.toSet());
+                selectedInfraValSet.add(new InfrastructureValueSet(infrastructureValueSet.getType(), selectedSet));
+            });
+            return selectedInfraValSet;
+        } else if (includes != null && !includes.isEmpty()) {
+            infrastructures.forEach(infrastructureValueSet -> {
+                Set<InfrastructureParameter> selectedSet = infrastructureValueSet.getValues().stream()
```
hasuniea: shall we assign infrastructureValueSet.getValues() to a variable, It will improve the readability?

common/src/main/java/org/wso2/testgrid/common/config/InfrastructureConfig.java
```diff
@@ -43,6 +43,8 @@
     private ContainerOrchestrationEngine containerOrchestrationEngine;
     private Properties parameters = new Properties();
     private List<Provisioner> provisioners;
+    private List<String> excludes;
```
pasindujw: IMO, something like "excludeInfraParams" might add more meaning since we are only excluding infra parameters. :)  WDYT?

core/src/test/java/org/wso2/testgrid/core/command/GenerateTestPlanCommandTest.java
```diff
@@ -116,7 +117,8 @@ public void testExecute(String jobConfigFile, String workingDir) throws Exceptio
         InfrastructureParameter param = new InfrastructureParameter("ubuntu_16.04", DefaultInfrastructureTypes
                 .OPERATING_SYSTEM, "{}", true);
         InfrastructureCombination comb1 = new InfrastructureCombination(Collections.singleton(param));
```
sameerawickramasekara: Shall we change the 'comb1' to a more meaningful name as well?

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -131,4 +136,44 @@ private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombinatio
         }
     }
 
+    /**
+     * This class provides set of filtered infrastructure values.
+     * <p>
+     * The filtering is executed by reading exclude and include attributes in the @{@link TestgridYaml} file.
+     * If exclude attribute is set in @{@link TestgridYaml} and if those are exists in the current infrastructure
```
azinneera: grammatical error "those are exists" -> "those exist"

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -131,4 +136,44 @@ private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombinatio
...
+     * value set, those will be eliminated.
+     * Further, if include attribute is set in @{@link TestgridYaml} and if those are exists in the current
+     * infrastructure value set, only those infrastructure will be consider when creating combinations.
```
azinneera: grammatical error "consider" -> "considered"

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -50,14 +54,15 @@
...
         if (logger.isDebugEnabled()) {
-            logger.debug("Retrieved value-set from database: " + valueSets);
+            logger.debug(String.format("Retrieved value-set from database: %s", valueSets));
         }
-        Set<InfrastructureCombination> infrastructureCombinations = getCombinations(valueSets);
+        Set<InfrastructureCombination> infrastructureCombinations = getCombinations(
+                filterInfrastructures(valueSets, testgridYaml));
         if (logger.isDebugEnabled()) {
-            logger.debug("Generated set of infrastructure combinations: " + infrastructureCombinations);
+            logger.debug(String.format("Generated set of infrastructure combinations: %s", infrastructureCombinations));
```
azinneera: Shall we have the string concatenation with `+` here? AFAIK, `String#format` is slower since it first parses the input with regular expressions and then fills in the parameters. Concatenation with `+` is optimized by javac and it uses StringBuilder#append

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
```diff
@@ -131,4 +136,44 @@ private void addSubpropertiesAsInfrastructureParameters(InfrastructureCombinatio
         }
     }
 
+    /**
+     * This class provides set of filtered infrastructure values.
+     * <p>
+     * The filtering is executed by reading exclude and include attributes in the @{@link TestgridYaml} file.
+     * If exclude attribute is set in @{@link TestgridYaml} and if those are exists in the current infrastructure
+     * value set, those will be eliminated.
+     * Further, if include attribute is set in @{@link TestgridYaml} and if those are exists in the current
```
azinneera: Same as the above comment.

lasanthaDLPDS: Fixed with e6479361dd8952f6b1efed92fdf164956691cf0e

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
lasanthaDLPDS: Fixed with e6479361dd8952f6b1efed92fdf164956691cf0e

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
lasanthaDLPDS: Fixed with e6479361dd8952f6b1efed92fdf164956691cf0e

core/src/test/java/org/wso2/testgrid/core/command/GenerateTestPlanCommandTest.java
lasanthaDLPDS: Fixed with e6479361dd8952f6b1efed92fdf164956691cf0e

common/src/main/java/org/wso2/testgrid/common/config/InfrastructureConfig.java
lasanthaDLPDS: Since Include and exclude attributes are defined under the infrastructureConfig: section it indicates that include and exclude attributes are related to infrastructure. Therefore lest's keep attribute names as this way. 

Further with this improvement infrastructure_parameter table has to be modified. That means we should modify infrastructure parameter names as following way.

>mysql ---->  mysql-5.7
>postgres ----> postgres-10.3
>centos ----> centos-7.4

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
lasanthaDLPDS: If you are looking for performance only I believe that using string concatenation is the most efficient approach. But IMO if we are looking for readability the String.format thing is the much clearer.  This code is executed when debug is enabled, therefore shall we use String.format instead of string concatenation. WDYT?  :)

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
lasanthaDLPDS: This method is executed when Test Grid generates test plans. In order to generate test plans, we can execute the following command and InfrastructureValueSet is constructed by reading infrastructure_parameter table.  Therefore infrastructureValueSet.getValues() method always return a value.
- generate-test-plan -p <product_name> --file <testgrid_yaml_path> --workingDir 
<working_directory_path>

core/src/main/java/org/wso2/testgrid/infrastructure/InfrastructureCombinationsProvider.java
lasanthaDLPDS: IMO there is no any benefit here though we have introduced a separate variable for infrastructureValueSet.getValues() as we are not reusing this again within the code. :)

PR #800, opened 2018-06-13
Option to persist test scenario results in S3 and download per-scenario

### LINKED ISSUE #795: Persist scenario test results/artifacts in S3 in correct build directory.
**Description:**
Currently JTL files and scenario test logs are uploaded to S3 as the same directory they were created in the workplace.
(ex:   <Product Name>/workspace/identity-test-integration/scenarios/scenario01/jmeter/ xxxxx.jtl )
But those has to be persist in the relevant build directory of the product.
readme - testgrid s3 structure revealed

Addition to that, the artifacts will needed to be in zipped format also in case a user need to download them.

**Suggested Labels:**
Bug

**Suggested Assignees:**

**Affected Product Version:**
m30

**Steps to reproduce:**

**Related Issues:**

pasindujw:
**Purpose**
Fixes #795 
Now the logs are persisted in S3 as below;

readme - testgrid s3 structure revealed

The user is also able to download per-scenario results as below;
result

**Security checks**

### DIFF
--- a/web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
+++ b/web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
@@ -22,6 +22,7 @@ import org.apache.hc.core5.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.common.TestCase;
+import org.wso2.testgrid.common.TestGridConstants;
 import org.wso2.testgrid.common.TestPlan;
 import org.wso2.testgrid.common.TestScenario;
 import org.wso2.testgrid.common.config.ConfigurationContext;
@@ -68,6 +69,12 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+
+import static org.wso2.testgrid.common.TestGridConstants.TESTGRID_COMPRESSED_FILE_EXT;
+import static org.wso2.testgrid.web.utils.Constants.RESPONSE_HEADER_CONTENT_DISPOSITION;
+import static org.wso2.testgrid.web.utils.Constants.RESPONSE_HEADER_FILE_NAME;
+import static org.wso2.testgrid.web.utils.Constants.RESPONSE_HEADER_VALUE_APPLICATION_ZIP;
+import static org.wso2.testgrid.web.utils.Constants.RESPONSE_HEADER_VALUE_ATTACHMENT;
 
 /**
  * REST service implementation of TestPlan.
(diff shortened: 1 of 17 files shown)


web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -394,4 +401,71 @@ private TestExecutionSummary getTestExecutionSummary(TestPlan testPlan) {
...
+
+    /**
+     * Returns the result (archive file) to the given scenario of a test plan.
+     *
+     * @param id test plan id
+     * @param scenarioDir scenario directory name
+     * @return archived file of the results
+     */
+    @GET
+    @Path("/result/{id}/{scenarioDir}")
```
kasunbg: Shall we say 'test-plan-id' to make it more meaningful? Then we don't have to look at the javadoc.. :)

kasunbg: What is the scenarioDir?? If this is a jargon we are going to use, we should document this somewhere.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -394,4 +401,71 @@ private TestExecutionSummary getTestExecutionSummary(TestPlan testPlan) {
...
+                return Response
+                        .ok(artifactDownloadable.getArtifactStream(bucketKey))
+                        .type(RESPONSE_HEADER_VALUE_APPLICATION_ZIP)
+                        .header(RESPONSE_HEADER_CONTENT_DISPOSITION, RESPONSE_HEADER_VALUE_ATTACHMENT + "; " +
+                                RESPONSE_HEADER_FILE_NAME + "=\" " + scenarioDir +
+                                TESTGRID_COMPRESSED_FILE_EXT + "\"")
+                        .build();
+            } else {
+                return Response.status(Response.Status.NOT_FOUND)
+                        .entity("Couldn't found result-artifacts in the remote location").build();
```
kasunbg: At the same time, we should log an error to testgrid.log mentioning this error condition. 
The log should have more context - things like product name, test-plan-id, scenario-id, the s3 result-artifacts location you were trying, etc.

web/src/main/java/org/wso2/testgrid/web/api/TestPlanService.java
```diff
@@ -394,4 +401,71 @@ private TestExecutionSummary getTestExecutionSummary(TestPlan testPlan) {
...
+            }
+        } catch (TestGridDAOException e) {
+            String msg = "Error occurred while fetching the TestPlan by id : '" + id + "' ";
```
kasunbg: Just the 'id' enough to debug this error when all you have is the log file without access to the testgrid runtime/s3 etc.? From product support perspective, that's all we have usually.

PR #826, opened 2018-06-26
Support to run generic test scripts in testgrid for intg tests

kasunbg:
**Purpose**
Testgrid need to be able to run product integration tests. For that purpose, we are going to get the testgrid to have,

1. a proper data-passing mechanism through the testgrid's test-plan pipeline (infra -> deploy -> tests), 
2. Support for running generic test scripts in contrast to being jmeter specific. (Deprecated JmeterTestExecutor!)
3. Provide job's properties as input arguments to scripts testgrid execute.

### DIFF
--- a/automation/src/main/java/org/wso2/testgrid/automation/reader/JMeterTestReader.java
+++ b/automation/src/main/java/org/wso2/testgrid/automation/reader/JMeterTestReader.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import org.wso2.testgrid.automation.Test;
 import org.wso2.testgrid.automation.TestAutomationException;
 import org.wso2.testgrid.automation.TestEngine;
+import org.wso2.testgrid.common.TestGridConstants;
 import org.wso2.testgrid.common.TestScenario;
 
 import java.io.File;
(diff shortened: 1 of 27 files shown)


automation/src/main/java/org/wso2/testgrid/automation/reader/JMeterTestReader.java
```diff
@@ -65,20 +62,23 @@
                 Test test = new Test(scenario.getName(), TestEngine.JMETER, scriptList, scenario);
 
                 scripts.stream()
-                        .filter(x -> x.endsWith(SHELL_SUFFIX) && x.contains(PRE_STRING))
```
sameerawickramasekara: These per scenario **_pre_** and **_post_** scripts are not being used now, hence IMO we could remove the TestReading logic from here WDYT?

kasunbg: +1. Can be handled outside of this pr.

kasunbg: https://github.com/wso2/testgrid/issues/830

PR #827, opened 2018-06-28
Apply Config change set

### LINKED ISSUE #465: Support config change-sets of scenario test scripts in Testgrid
**Description:**
Scenario test labs team is introducing config value-sets for their scenarios. We need to apply each of these value-sets and run all the scenarios against these changes.

The function of these config-value-sets is to introduce delta changes to product configurations in the deployment. The scenario scripts need to be run against all these delta changes. 

Rather than _config-value-sets_, we can call them as _config-deltas_ as well. That is more meaningful I guess. :) 

**Affected Product Version:**
0.9.0-m11

N/A

madushadhanushka:
**Purpose**
Resolves #465 
Apply config change set on the product before running each of the test scenarios.

**Approach**
As the first step application read the testgrid.yaml file and read available configchangesets. Then it generate test plan for each of the test scenarios along with each or the config change set. You can assume config change set as fourth dimension of the test grid application. All the different config change sets iterate through all the different test scenario and generate test plan yaml file and database records.

Second step, When tester run execute test plan, then it send shell command to the relevant agent to download config change set script file from the Github as defined in job-config.yaml file "configChangeSetRepository" property. After it download the file, it execute the relevant shell command on agent node, then run test scenario and execute revert config change set script.

Once test scenarios executed It feed execution result into the test_scenario table in testgriddb database.

**Special Note**

- In order to use this feature with existing database you should update database with following SQL command.
`ALTER TABLE test_scenario ADD config_change_set_name varchar(255), ADD config_change_set_description varchar(255);`

- Need to define the config change set repository value in job-config.yaml file configChangeSetRepository key.
ex `configChangeSetRepository: https://github.com/wso2-incubator/identity-test-integration`

- Add configChangeSetBranchName to the job-config.yaml file to set the branch name for the config change set repository. ex `configChangeSetBranchName: master`

- PRODUCT_HOME environmental variable should be defined in jenkins property content in order to provide product home location. ex `PRODUCT_HOME=/opt/wso2/wso2is-5.5.0`

- Tinkerer API endpoint should be defined in config.properties file in DEPLOYMENT_TINKERER_REST_BASE_PATH key.

- Test scenarios with configchangeset will be generated by this application only if configchangeset defined in testgrid.yaml file. Otherwise, test scenarios will be execute same as previously without configchangeset combinations. 

configChangeSet key in testgrid.yaml contain two sub keys as "name" and "description". "name" key should be same as the  configchangeset script directory name and "description" contain small description what its about.


common/src/main/java/org/wso2/testgrid/common/TestScenario.java
```diff
@@ -75,6 +75,9 @@
     @Column(name = "is_pre_script_success")
     private boolean isPreScriptSuccessful = false;
 
+    @Column(name = "configChangeSet_name")
```
kasunbg: Why mix camelCase and underscore?

common/src/main/java/org/wso2/testgrid/common/config/JobConfigFile.java
```diff
@@ -85,6 +86,14 @@ public void setScenarioTestsRepository(String scenarioTestsRepository) {
         this.scenarioTestRepository = scenarioTestsRepository;
     }
 
+    public String getConfigChangeSetRepository() {
```
kasunbg: Missing docs.

core/pom.xml
```diff
@@ -140,5 +140,17 @@
             <groupId>org.apache.velocity</groupId>
             <artifactId>velocity</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
```
kasunbg: We are already using gson afair. Shall we use that instead?

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+        try {
+            URL configChangeSetRepoPath = new URL(testPlan.getConfigChangeSetRepository());
+            String filePath = configChangeSetRepoPath.getPath();
+            String fileName = filePath.substring(filePath.lastIndexOf('/') + 1);
+            String[] initShellCommand = {
+                    "mkdir repos",
+                    "cd repos && curl -LJO " +
+                            testPlan.getConfigChangeSetRepository() + "/archive/master.tar.gz   &>/dev/null",
+                    "cd repos && tar xvzf " + fileName + "-master.tar.gz  &>/dev/null",
+                    "chmod -R 755 repos/" + fileName + "-master/config-sets/ &>/dev/null"
```
kasunbg: This will fail in windows env.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+            };
+            return applyShellCommandOnAgent(testPlan, initShellCommand);
+        } catch (MalformedURLException e) {
+            logger.warn("Error parsing scenario path " + e);
```
kasunbg: This will swallow the exception stack trace. `e` should be provided as the second argument. Also, better if we can print which URL failed.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+
+    /**
+     * Apply config change set script before and after run test scenarios
+     *
+     * @param testPlan the test plan
+     * @param configChangeSet   config change set
+     * @param isInit run apply config-script if true. else, run revert-config script
+     * @return execution passed or failed
+     */
+    private boolean applyConfigChangeSet(TestPlan testPlan, ConfigChangeSet configChangeSet, boolean isInit) {
```
kasunbg: Shall we hand-over the handling of configChangeSets to its own class? That'll improve the readability a lot imo.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+                        for (String shellCommand : initShellCommand) {
+                            logger.info("exec " + shellCommand);
+                            Content shellCommandResponse = sendShellCommand(tinkererHost, authenticationToken
+                                    , agent, shellCommand);
+                        }
+                    }
+                }
+                return true;
+            } catch (JSONException e) {
+                logger.warn("JSON parsing error " + e);
```
kasunbg: Same as above.

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -487,12 +494,37 @@ private Yaml createYamlInstance() {
...
+                             newTestScenario.setDescription(testScenario.getDescription());
+                             newTestScenario.setDir(testScenario.getDir());
+                             newTestScenario.setIsPostScriptSuccessful(testScenario.isPostScriptSuccessful());
+                             newTestScenario.setIsPreScriptSuccessful(testScenario.isPreScriptSuccessful());
+                             newTestScenario.setStatus(testScenario.getStatus());
+                             newTestScenario.setTestPlan(testScenario.getTestPlan());
+                             newTestScenario.setConfigChangeSetName(configChangeSet.getName());
+                             newTestScenario.setSummaryGraphs(testScenario.getSummaryGraphs());
+                             newTestScenario.setTestCases(testScenario.getTestCases());
+                             modifiedTestScenarios.add(newTestScenario);
```
kasunbg: Shouldn't you remove the old test scenarios first?

common/src/main/java/org/wso2/testgrid/common/TestScenario.java
madushadhanushka: I will change it to "config_change_set_name". Any better suggestion?

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
madushadhanushka: Records are not persisted till in this line. It persist only after this step. 
Therefore, it is ok.
Problem here it already assign value for ID with the beginning of the parsing testgrid.yaml file. Therefore I had to create new ArrayList and then persist it.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
harshanL: +1

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+
+    /**
+     * Revert back changes did in initConfigChangeSet
+     *
+     * @param testPlan  The test plan
+     * @return          True if execution success. Else, false
+     */
+    private boolean deInitConfigChangeSet(TestPlan testPlan) {
+        String[] deInitShellCommand = {
+                "rm -rf repos"
```
harshanL: This will also fail in windows env.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -339,6 +339,177 @@ private void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploy
...
+        try {
+            Content agentResponse = Request.Get(tinkererHost + "/agents")
+                    .addHeader(HttpHeaders.AUTHORIZATION, authenticationToken).
+                            execute().returnContent();
+            try {
+                // Todo use with Agent class with jackson mapper
+                JSONArray agentArray = new JSONArray(agentResponse.asString());
+                for (int index = 0; index < agentArray.length(); index++) {
+                    JSONObject agent = agentArray.getJSONObject(index);
+                    if (agent.getString("testPlanId").equals(testPlan.getId())) {
```
harshanL: Is it guaranteed to return an not null value even if agent.getString("testPlanId") does not exists?

azinneera:
The PR should refer the issue that is being resolved. i.e. Resolves #<issue-number>

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -294,39 +298,38 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
-                            logger.error(String.format("Error parsing the results for the" +
-                                    " scenario %s", testScenario.getName()));
+            List<ConfigChangeSet> configChangeSetList = testPlan.getScenarioConfig().getConfigChangeSets();
+            // Run test with config change sets
+            if (configChangeSetList != null) {
+                OSCategory osCategory = getOSCatagory(testPlan.getInfraParameters());
+                Optional<ConfigChangeSetExecutor> configChangeSetExecutor =
+                        ConfigChangeSetFactory.getExecuter(osCategory);
+                if (configChangeSetExecutor.isPresent()) {
+
```
azinneera: Unnecessary new line

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetFactory.java
```diff
@@ -0,0 +1,53 @@
...
+ * @since 1.0.0
+ */
+public class ConfigChangeSetFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger(TinkererClientFactory.class);
+
+    /**
+     * Generate Config change set executor for given OS
+     * @param osCategory
+     * @return
```
azinneera: Incomplete javadoc comment

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -299,39 +303,48 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
-                                "scenario %s", testScenario.getName()));
+                        // Remove config set repos on agent
+                        configChangeSetExecutor.get().deInitConfigChangeSet(testPlan);
                     }
-                } catch (ScenarioExecutorException e) {
-                    // we need to continue the rest of the tests irrespective of the exception thrown here.
-                    logger.error(StringUtil.concatStrings("Error occurred while executing the SolutionPattern '",
-                            testScenario.getName(), "' in TestPlan\nCaused by "), e);
+                } else {
+                    logger.error("Unable to find a Tinker Executor for OS category " + osCategory);
```
pasindujw: Shall we add more details about the situation such as testplan-id, infra-combination, etc?

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -340,6 +353,44 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
+        try {
+            scenarioExecutor.execute(testScenario, deploymentCreationResult, testPlan);
+            Optional<ResultParser> parser = ResultParserFactory.getParser(testPlan, testScenario);
+            if (parser.isPresent()) {
+                try {
+                    ResultParser resultParser = parser.get();
+                    resultParser.parseResults();
+                    resultParser.archiveResults();
+                } catch (ResultParserException e) {
+                    logger.error(String.format("Error parsing the results for the" +
```
pasindujw: lets include the ResultParserException (e) in the end of the log as well.

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -498,12 +505,38 @@ private Yaml createYamlInstance() {
...
                         combination.getParameters());
                 testPlan.getInfrastructureConfig().setParameters(configAwareInfraCombination);
+
+                ScenarioConfig scenarioConfig = testgridYaml.getScenarioConfig();
+                List<ConfigChangeSet> configChangeSets = scenarioConfig.getConfigChangeSets();
+                if (configChangeSets != null) {
+                    List<TestScenario> modifiedTestScenarios = new ArrayList<>();
+                    for (ConfigChangeSet configChangeSet : configChangeSets) {
+                        for (TestScenario testScenario : scenarioConfig.getScenarios()) {
+                             TestScenario newTestScenario = new TestScenario();
```
pasindujw: Since the array is calling 'modifiedTestScenarios', I think `modifiedTestScenario` would be a better name, WDYT?

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -340,6 +353,44 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
+        } catch (ScenarioExecutorException e) {
+            // we need to continue the rest of the tests irrespective of the exception thrown here.
+            logger.error(StringUtil.concatStrings("Error occurred while executing the SolutionPattern '",
+                    testScenario.getName(), "' in TestPlan\nCaused by "), e);
+        }
+        try {
+            persistTestScenario(testScenario);
+        } catch (TestPlanExecutorException e) {
+            logger.error(StringUtil.concatStrings(
+                    "Error occurred while persisting test scenario ", testScenario.getName()), e);
```
pasindujw: Do we need to throw the exception to the object whom calling this?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+            List<String> initShellCommand = new ArrayList<>();
+            initShellCommand.add("mkdir repos");
+            initShellCommand.add("cd repos && curl -LJO " +
+                    testPlan.getConfigChangeSetRepository() + "/archive/master.tar.gz   &>/dev/null");
+            initShellCommand.add("cd repos && tar xvzf " + fileName + "-master.tar.gz  &>/dev/null");
+            initShellCommand.add("chmod -R 755 repos/" + fileName + "-master/config-sets/ &>/dev/null");
+            return applyShellCommandOnAgent(testPlan, initShellCommand);
+        } catch (MalformedURLException e) {
+            logger.warn("Error parsing scenario repository path for ".
+                    concat(testPlan.getConfigChangeSetRepository()), e);
```
pasindujw: Do we need to throw exception to the calling object?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Revert back changes did in initConfigChangeSet
+     *
+     * @param testPlan  The test plan
+     * @return          True if execution success. Else, false
+     */
+    @Override
+    public boolean deInitConfigChangeSet(TestPlan testPlan) {
```
pasindujw: I think `revertConfigChangeSet` would sounds better, WDYT?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+        List<String> deInitShellCommand = new ArrayList<>();
+        deInitShellCommand.add("rm -rf repos");
+        return applyShellCommandOnAgent(testPlan, deInitShellCommand);
+    }
+
+
```
pasindujw: Additional new line.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+                ConfigurationContext.ConfigurationProperties.DEPLOYMENT_TINKERER_REST_BASE_PATH);
+        // Get authentication detail from config.properties and encode with BASE64
+        String authenticationString = ConfigurationContext.getProperty(
+                ConfigurationContext.ConfigurationProperties.DEPLOYMENT_TINKERER_USERNAME) + ":" +
+                ConfigurationContext.getProperty(
+                        ConfigurationContext.ConfigurationProperties.DEPLOYMENT_TINKERER_PASSWORD);
+        String authenticationToken = "Basic " + Base64.getEncoder().encodeToString(
+                authenticationString.getBytes(StandardCharsets.UTF_8));
+        boolean executionStatus = true;
+        String agentLink = tinkererHost + "test-plan/" + testPlan.getId() + "/agents";
```
pasindujw: Shall we use Paths (https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html) to derive the URL?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+                        code = result.get("code").getAsString();
+                    }
+                    if (result.get("response") != null) {
+                        response = result.get("response").getAsString();
+                    }
+                    logger.info("Agent exit value: " + exitValue);
+                    logger.info("Agent code: " + code);
+                    logger.info("Agent response: \n" + response);
+                    // Agent code default is SHELL, if time out then, it return code as 408
+                    if (code.equals("408") || exitValue != 0) {
```
pasindujw: Shall we refer the HTTP status code as a constant from a referable class?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+                        executionStatus = false;
+                        logger.error("Agent script execution failed with code " + code + " exit value " + exitValue);
+                    }
+                }
+            }
+            return executionStatus;
+        } catch (ScenarioExecutorException e) {
+            logger.error("Error in API call request to execute script ".concat(tinkererHost), e);
```
pasindujw: Can we include the agent-id in the log?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,239 @@
...
+     * @param authenticationKey     Tinkerer authentication key as BASE64 encoded string
+     * @param agent                 Agent details
+     * @param script                Script that need to be executed
+     * @return                      Shell execution output
+     * @throws ScenarioExecutorException          Request post exception
+     */
+    private Content sendShellCommand(String tinkererHost, String authenticationKey, Agent agent, String script)
+            throws ScenarioExecutorException {
+        Content result;
+        String requestLink = tinkererHost + "test-plan/" + agent.getTestPlanId() +
```
pasindujw: Use Paths to derive the URL.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorWindows.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Initialize agent before running config change set
+     *
+     * @param testPlan  The test plan
+     * @return          True if execution success. Else, false
+     */
+    @Override
+    public boolean initConfigChangeSet(TestPlan testPlan) {
+        return false;
```
pasindujw: Lets add a `todo` comment here saying that this has to be implemented. :)

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Optional;
+
+/**
+ * This factory create appropriate config change set executor class for given OS
+ *
+ * @since 1.0.0
+ */
+public class ConfigChangeSetFactory {
+
+    private static final Logger logger = LoggerFactory.getLogger(TinkererClientFactory.class);
```
pasindujw: The logger class has to be 'ConfigChangeSetFactory'

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
madushadhanushka: Fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorWindows.java
madushadhanushka: Added into the beginning of the class

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetFactory.java
madushadhanushka: fixed

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetFactory.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ConfigChangeSetFactory.class);
+
+    /**
+     * Generate Config change set executor for given OS
+     * @param osCategory type of agent OS as WINDOWS or UNIX
+     * @return Object to run config change set commands
+     */
+    public static Optional<ConfigChangeSetExecutor> getExecuter(TestPlanExecutor.OSCategory osCategory) {
+        if (osCategory.equals(TestPlanExecutor.OSCategory.UNIX)) {
+            logger.info("OSCatagory is UNIX");
```
kasunbg: We can make this a debug log.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -299,39 +303,49 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
-                            resultParser.archiveResults();
-                        } catch (ResultParserException e) {
-                            logger.error(String.format("Error parsing the results for the" +
-                                    " scenario %s", testScenario.getName()));
+            List<ConfigChangeSet> configChangeSetList = testPlan.getScenarioConfig().getConfigChangeSets();
+            // Run test with config change sets
+            if (configChangeSetList != null) {
+                OSCategory osCategory = getOSCatagory(testPlan.getInfraParameters());
+                Optional<ConfigChangeSetExecutor> configChangeSetExecutor =
+                        ConfigChangeSetFactory.getExecuter(osCategory);
```
kasunbg: s/getExecuter/getExecutor/g

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
```diff
@@ -299,39 +303,49 @@ public void runScenarioTests(TestPlan testPlan, DeploymentCreationResult deploym
...
+                                            deploymentCreationResult, true)) {
+                                // Run test scenarios for relevant config change set
+                                for (TestScenario testScenario : testPlan.getTestScenarios()) {
+                                    if (configChangeSet.getName().equals(testScenario.getConfigChangeSetName())) {
+                                        executeTestScenario(testScenario, deploymentCreationResult, testPlan);
+                                    }
+                                }
+                                // Revert config change set after running test scenarios
+                                // If revert back is not success then, break all test scenarios running
+                                if (!configChangeSetExecutor.get().applyConfigChangeSet(testPlan, configChangeSet,
```
kasunbg: IMO, we need to have two separate methods to handle #apply and #revert scenarios.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+    @Override
+    public boolean initConfigChangeSet(TestPlan testPlan) {
+        try {
+            URL configChangeSetRepoPath = new URL(testPlan.getConfigChangeSetRepository());
+            String filePath = configChangeSetRepoPath.getPath();
+            String fileName = filePath.substring(filePath.lastIndexOf('/') + 1);
+            List<String> initShellCommand = new ArrayList<>();
+            initShellCommand.add("mkdir repos");
+            initShellCommand.add("cd repos && curl -LJO " +
+                    testPlan.getConfigChangeSetRepository() + "/archive/master.tar.gz   &>/dev/null");
```
kasunbg: Git branch name is hard-coded. It won't be master always.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+        // Get authentication detail from config.properties and encode with BASE64
+        String authenticationString = ConfigurationContext.getProperty(
+                ConfigurationContext.ConfigurationProperties.DEPLOYMENT_TINKERER_USERNAME) + ":" +
+                ConfigurationContext.getProperty(
+                        ConfigurationContext.ConfigurationProperties.DEPLOYMENT_TINKERER_PASSWORD);
+        String authenticationToken = "Basic " + Base64.getEncoder().encodeToString(
+                authenticationString.getBytes(StandardCharsets.UTF_8));
+        boolean executionStatus = true;
+        String agentLink = tinkererHost + "test-plan" + TestGridConstants.FILE_SEPARATOR + testPlan.getId() +
+                TestGridConstants.FILE_SEPARATOR + "agents";
```
kasunbg: What you are building here is a URL. URLs has "/" forward-slash as the separator always, irrespective of the OS. 
You could rather use java.net.URL to build the agentLink url.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+     * @param script                Script that need to be executed
+     * @return                      Shell execution output
+     * @throws ScenarioExecutorException          Request post exception
+     */
+    private Content sendShellCommand(String tinkererHost, String authenticationKey, Agent agent, String script)
+            throws ScenarioExecutorException {
+        Content result;
+        String requestLink = tinkererHost + "test-plan" + TestGridConstants.FILE_SEPARATOR + agent.getTestPlanId() +
+                TestGridConstants.FILE_SEPARATOR + "agent" + TestGridConstants.FILE_SEPARATOR + agent.getInstanceName()
+                + TestGridConstants.FILE_SEPARATOR + "operation";
```
kasunbg: Same as above.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+        try {
+            result = Request.Post(requestLink).addHeader(HttpHeaders.AUTHORIZATION, authenticationKey)
+                    .addHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+                    .bodyString("{\"request\":\"" + script + "\",\"code\":\"SHELL\"}", ContentType.APPLICATION_JSON).
+                            execute().returnContent();
+        } catch (IOException e) {
+            throw new ScenarioExecutorException(StringUtil.concatStrings(
```
kasunbg: This should be a ConfigChangesetExecutorException, isn't it?

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+            initShellCommand.add("mkdir repos");
+            initShellCommand.add("cd repos && curl -LJO " +
+                    testPlan.getConfigChangeSetRepository() + "/archive/master.tar.gz   &>/dev/null");
+            initShellCommand.add("cd repos && tar xvzf " + fileName + "-master.tar.gz  &>/dev/null");
+            initShellCommand.add("chmod -R 755 repos/" + fileName + "-master/config-sets/ &>/dev/null");
+            return applyShellCommandOnAgent(testPlan, initShellCommand);
+        } catch (MalformedURLException e) {
+            logger.warn("Error parsing scenario repository path for ".
+                    concat(testPlan.getConfigChangeSetRepository()), e);
+            return false;
```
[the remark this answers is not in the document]
kasunbg: This is true for applyConfigSet as well. If it failed, we should try to revert, and retry periodically for fixed number of times.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                    if (result.get("code") != null) {
+                        code = result.get("code").getAsString();
+                    }
+                    if (result.get("response") != null) {
+                        response = result.get("response").getAsString();
+                    }
+                    logger.info("Agent exit value: " + exitValue);
+                    logger.info("Agent code: " + code);
+                    logger.info("Agent response: \n" + response);
```
kasunbg: Above logs are missing the context. When you have a log file with thousands of lines, the context becomes important. In here, we should print what's the testplan and agent, which config set you are trying to apply etc. 

We can merge these into one log line as well imo.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+                    // Agent code default is SHELL, if time out then, it return code as 408
+                    if (code.equals(String.valueOf(HttpStatus.SC_REQUEST_TIMEOUT)) || exitValue != 0) {
```
kasunbg: What if the code is HTTP 500? Is it enough to only check for request timeout? 

As a general rule in HTTP status codes, 2xx codes are considered successful, 4xx are client-side errors, and 5xx are server-side errors. In this case, any code other than 2xx code range is an error scenario.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -0,0 +1,242 @@
...
+                        executionStatus = false;
+                        logger.error("Agent script execution failed with code " + code + " exit value " + exitValue);
```
kasunbg: missing ctxt.

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorWindows.java
kasunbg: Yes, we should add a todo.. This is a must have for this task to be done-done though.

dao/src/main/java/org/wso2/testgrid/dao/uow/TestPlanUOW.java
```diff
@@ -61,6 +61,7 @@ public TestPlan persistTestPlan(TestPlan testPlan) throws TestGridDAOException {
             persisted.setScenarioTestsRepository(testPlan.getScenarioTestsRepository());
             persisted.setInfrastructureRepository(testPlan.getInfrastructureRepository());
             persisted.setDeploymentRepository(testPlan.getDeploymentRepository());
+            persisted.setConfigChangeSetRepository(testPlan.getConfigChangeSetRepository());
```
kasunbg: We need to do this for TestPlan#clone operation as well - https://github.com/wso2/testgrid/blob/aa462b763be00a4a8169e39daaa1152fc184d77d/common/src/main/java/org/wso2/testgrid/common/TestPlan.java#L381

kasunbg:
This is a good PR with good code quality best practices. :+1:

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
madushadhanushka: This is the timeout response from the agent. Current implementation it only set back timeout status as code back to the test executor. Other responses are not handle. This will be fixed with next feature development.

kasunbg: Well, timeout status code may be the only thing hard-coded. But, other error codes can also be present for numerous reasons. For example, if an exception get thrown by the agent rest service, then the tinkerer will just receive a 500 status code. Another example, say the API contract of the agent rest service changed. Then, your HTTP requests may fail with a 404 not found. 
Therefore, it's important to handle other status codes.
We can of course do this in a separate PR, but please open a github ticket with highest priority. :)

core/src/main/java/org/wso2/testgrid/core/configchangeset/ConfigChangeSetExecutorUnix.java
```diff
@@ -171,11 +171,19 @@ public boolean initConfigChangeSet(TestPlan testPlan) {
             String filePath = configChangeSetRepoPath.getPath();
             String fileName = filePath.substring(filePath.lastIndexOf('/') + 1);
             List<String> initShellCommand = new ArrayList<>();
+            // set default branch as master if not mention in testgrid.yaml
+            String branchName = "master";
+            if (testPlan.getConfigChangeSetBranchName() != null) {
+                branchName = testPlan.getConfigChangeSetBranchName();
+            } else {
+                logger.warn("Config change set repository branch name is not set. " +
+                        "Using default repository branch as master");
```
kasunbg: Missing ctxt.

core/src/main/java/org/wso2/testgrid/core/TestPlanExecutor.java
kasunbg: Catching exceptions and logging them is not considered a bad practice!

Issue #849, opened 2018-07-04
Check the possibility of making AWS stack deletion non-blocking


yasassri:
**Description:**

ATM we wait till the stack deletion is completed which takes around 10mins most of the time. If we can handover the stack deletion to AWS and continue this time can be saved.

PR #859, opened 2018-07-05
Make stack deletion waiting configurable

yasassri:
**Purpose**

Make stack deletion waiting configurable. Resolves https://github.com/wso2/testgrid/issues/849

### DIFF
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
@@ -304,15 +304,20 @@ public class AWSProvider implements InfrastructureProvider {
         DeleteStackRequest deleteStackRequest = new DeleteStackRequest();
         deleteStackRequest.setStackName(stackName);
         stackdestroy.deleteStack(deleteStackRequest);
-        logger.info(StringUtil.concatStrings("Waiting for stack : ", stackName, " to delete.."));
-        Waiter<DescribeStacksRequest> describeStacksRequestWaiter = new
-                AmazonCloudFormationWaiters(stackdestroy).stackDeleteComplete();
-        try {
-            describeStacksRequestWaiter.run(new WaiterParameters<>(new DescribeStacksRequest()
-                    .withStackName(stackName)));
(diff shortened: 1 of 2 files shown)


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -304,15 +304,21 @@ private boolean doRelease(InfrastructureConfig infrastructureConfig, String stac
...
-            describeStacksRequestWaiter.run(new WaiterParameters<>(new DescribeStacksRequest()
-                    .withStackName(stackName)));
-        } catch (WaiterUnrecoverableException e) {
-            throw new TestGridInfrastructureException("Error occured while waiting for Stack :"
-                    + stackName + " deletion !");
+        logger.info(StringUtil.concatStrings("Stack : ", stackName, " is handed over for deletion!"));
+
+        boolean waitForStackDeletion = Boolean.parseBoolean(ConfigurationContext.getProperty(ConfigurationContext.
+                ConfigurationProperties.WAIT_FOR_STACK_DELETION));
+        if (waitForStackDeletion) {
```
kasunbg: What if deletion failed? How can we track that? Would it make sense to run the waiting task asynchronously?

yasassri: ATM if the deletion fails we do not do anything. Yes we can do an asynchronous check and print a log if stack deletion fails. Another approach would be to maintain the states of stacks created, say when deletion is handed over we add an DB entry, later we can introduce a separate task to remove any dangling stacks. 
As a side note, are we going to use this implementation in the future? If the we are just relying on infra.sh we shouldn't invest much on this implementation.

kasunbg: I think we must handle deletion failures. When we have 50+ concurrent builds, we'll have same number of cfn stacks. Manually managing this is not an option. 

We can track how to handle the stack deletion failure in a separate ticket.

On your second point, even with infra.sh, we need some awareness that we are running a cloudformation script IMO. This is needed to provide specialized support such as determining the region to use when doing parallel builds.

PR #882, opened 2018-07-12
Handle creation of empty executors when parallel executor count is larger than test-plan count

yasassri:
This PR handles creation of empty executors when parallel executor count is larger than test-plan count.

### DIFF
--- a/jenkins/pipelines/parallel-execution/Jenkinsfile
+++ b/jenkins/pipelines/parallel-execution/Jenkinsfile
@@ -5,6 +5,17 @@ def loadProperties() {
         properties = readProperties file: "${TESTGRID_HOME}/config.properties"
     }
 }
+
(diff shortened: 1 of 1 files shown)


jenkins/pipelines/parallel-execution/Jenkinsfile
```diff
@@ -90,13 +85,13 @@ pipeline {
                 // Clone scenario repo
                 sh "mkdir -p ${SCENARIOS_LOCATION}"
                 dir("${SCENARIOS_LOCATION}"){
-                    git branch: 'master', url: "${SCENARIOS_REPOSITORY}"
+                    git branch: 'grid2', url: "${SCENARIOS_REPOSITORY}"
```
kasunbg: shall we keep master as the branch? :)

PR #886, opened 2018-07-15
 Improve email content to show error/fail test cases 

kasunbg:
**Purpose**

Previously, we showed unprocessed file contents in Test-TestSuite.txt.
Instead, now we process the TEST-TestSuite.xml file.
This is the surefire report file that contain a summary of all the test cases
in a parseable xml format.


reporting/src/main/java/org/wso2/testgrid/reporting/EmailReportProcessor.java
```diff
@@ -69,76 +70,73 @@ public EmailReportProcessor() {
...
+    public List<TPResultSection> generatePerTestPlanSection(Product product, List<TestPlan> testPlans)
             throws ReportingException {
-        List<TestPlanResultSection> perTestPlanMap = new ArrayList<>();
+        List<TPResultSection> perTestPlanList = new ArrayList<>();
         String testGridHost = ConfigurationContext.getProperty(ConfigurationContext.
                 ConfigurationProperties.TESTGRID_HOST);
         String productName = product.getName();
         for (TestPlan testPlan : testPlans) {
             if (testPlan.getStatus().equals(Status.SUCCESS)) {
+                logger.debug(String.format("Test plan ,%s, status is set to success.Not including in email report. "
```
yasassri: Should this be a info log?

reporting/src/main/java/org/wso2/testgrid/reporting/EmailReportProcessor.java
```diff
@@ -69,76 +70,73 @@ public EmailReportProcessor() {
...
-            try {
-                testPlanResultSection.setResult(getIntegrationTestLogResult(testPlan));
-            } catch (ReportingException e) {
-                logger.error("Integration test results of the test-plan with id " + testPlan.getId() +
-                        " is not added to the email-report due to a reporting exception occurred.", e.getMessage());
+            final String infraCombination = testPlan.getInfraParameters();
+            final String dashboardLink = String.join("/", testGridHost, productName, deploymentPattern,
+                    TEST_PLANS_URI, testPlanId);
+
+            SurefireReporter surefireReporter = new SurefireReporter();
```
pasindujw: We can take the SurefireReporter out of the loop of test-plans and avoid creating objects per-loop-turn

reporting/src/main/java/org/wso2/testgrid/reporting/EmailReportProcessor.java
kasunbg: +1

reporting/src/main/java/org/wso2/testgrid/reporting/surefire/TestResult.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.List;
+
+/**
+ * The summarized surefire test results.
+ */
+public class TestResult {
+    String totalTests = "?";
+    String totalFailures = "?";
+    String totalErrors = "?";
+    String totalSkipped = "?";
```
charithag: Access specifier can be private. Also since this is holds a count as a numeric value and count could not be less than 0, can't we declare these counts as int and use -1 to indicate values not initialized?

kasunbg: Access modifier is set as package-private because these are accessed by SurefireReporter class.

The variable type is a string because the underlying maven-surefire-reporter dependency actually returns these as strings (ref: https://github.com/apache/maven-surefire/blob/master/surefire-report-parser/src/main/java/org/apache/maven/plugins/surefire/report/SurefireReportParser.java#L152).
Since we do not need this as an int, I kept the String type as it is.

Question mark, ?, is set because we are going to display this in the email report. We can opt for a "-1" string as well here.

reporting/src/main/java/org/wso2/testgrid/reporting/EmailReportProcessor.java
```diff
@@ -69,76 +69,76 @@ public EmailReportProcessor() {
...
+        displayCount = displayCount < maxDisplayTestCount ? displayCount : maxDisplayTestCount;
+
+        tests = new ArrayList<>(tests.subList(0, displayCount));
+        if (displayCount < actualCount) {
+            TestResult.TestCaseResult continuation = new TestResult.TestCaseResult();
+            continuation.setClassName("...");
+            TestResult.TestCaseResult allTestsInfo = new TestResult.TestCaseResult();
+            allTestsInfo.setClassName("(view complete list of tests (" + actualCount + ") in testgrid-live..)");
+            tests.add(continuation);
+            tests.add(continuation);
```
harshanL: Duplicated line. Shall we remove that?

kasunbg: This is done by purpose. I wanted the email to contain two '----' lines if the number of test cases are trimmed.

PR #887, opened 2018-07-15
Improve latest test plans retrieving query

lasanthaDLPDS:
**Purpose**
The purpose of this PR is to modify the latest test plans retrieving query. 

**Goals**
When retrieving latest test plans for a given product, as per the current implementation, it is mainly considered the last modified timestamp of the test plan table and get the latest test plans. Since we stored the test plan number in the DB, it is more appropriate if we consider the maximum of the test plan number for an infra combination and a deployment pattern when we querying the latest test plans for a given product. Hence modified the query in this way. 

**Security checks**

**Test environment**
- JDK - ORACLE 1.8
- OS - UBUNTU 18.04
- DB- MySQL 5.7

### DIFF
--- a/core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
+++ b/core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
@@ -28,8 +28,10 @@ import org.wso2.testgrid.common.TestGridConstants;
 import org.wso2.testgrid.common.TestPlan;
 import org.wso2.testgrid.common.TestScenario;
 import org.wso2.testgrid.common.config.DeploymentConfig;
+import org.wso2.testgrid.common.config.InfrastructureConfig;
 import org.wso2.testgrid.common.config.InfrastructureConfig.Provisioner;
 import org.wso2.testgrid.common.config.JobConfigFile;
+import org.wso2.testgrid.common.config.ScenarioConfig;
 import org.wso2.testgrid.common.config.Script;
 import org.wso2.testgrid.common.config.TestgridYaml;
 import org.wso2.testgrid.common.exception.CommandExecutionException;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -180,8 +180,11 @@ private void processTestgridConfiguration(String jobConfigFilePath)
     }
 
     private void processTestgridYaml(TestgridYaml testgridYaml, JobConfigFile jobConfigFile)
-            throws CommandExecutionException, IOException, TestGridDAOException {
-        // TODO: validate the testgridYaml. It must contain at least one infra provisioner, and one scenario.
+            throws CommandExecutionException, TestGridDAOException {
+        if (!validateTestgridYaml(testgridYaml)) {
+            throw new CommandExecutionException(
+                    "Invalid tesgridYaml file is found,Please verify the content of the testgridYaml file");
```
harshanL: Shall we break this comment into two lines instead of separating using a ',' ?

kasunbg:
@charithag this is another accuracy issue we found. We have a method to get latest-test-plans of a given product via the DB. This output was incorrect when there were parallel builds happening in two or more slaves.

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -628,4 +631,35 @@ private Path getTestGridYamlLocation(String directory) {
...
+        Boolean isValidTestgridYaml = true;
+        InfrastructureConfig infrastructureConfig = testgridYaml.getInfrastructureConfig();
+        ScenarioConfig scenarioConfig = testgridYaml.getScenarioConfig();
+        if (infrastructureConfig != null) {
+            if (infrastructureConfig.getProvisioners().isEmpty()) {
+                logger.error("testgrid.yaml doesn't contain at least single infra provisioner, Invalid testgrid.yaml");
+                isValidTestgridYaml = false;
+            }
+        } else {
+            logger.error("testgrid.yaml doesn't have defined the infra configuration, Invalid testgrid.yaml");
```
kasunbg: The decision of logging this is upto the caller of #validateTestgridYaml method. The validateTestgridYaml method's sole purpose is to check the validity and return true/false imo.

dao/src/main/java/org/wso2/testgrid/dao/repository/TestPlanRepository.java
```diff
@@ -195,16 +195,17 @@ public TestPlan getLastBuild(Product product) {
...
-                + "tp.DEPLOYMENTPATTERN_id=dp.id  and tp.DEPLOYMENTPATTERN_id in "
-                + "(select id from deployment_pattern where PRODUCT_id= ? ) group by tp.infra_parameters,dp.id) as x "
-                + "on t.infra_parameters=x.infra_parameters AND t.modified_timestamp=x.time AND "
-                + "t.DEPLOYMENTPATTERN_id=x.dep_id;";
+        String sql = "select tp.* from test_plan tp inner join (Select distinct infra_parameters, max(test_run_number) "
+                + "as test_run_number from test_plan where  DEPLOYMENTPATTERN_id in (select id from deployment_pattern "
+                + "where PRODUCT_id= ?) group by infra_parameters) as latest_test_run_nums on "
+                + "tp.infra_parameters=latest_test_run_nums.infra_parameters and "
+                + "tp.test_run_number=latest_test_run_nums.test_run_number and tp.DEPLOYMENTPATTERN_id in "
+                + "(select id from deployment_pattern where PRODUCT_id= ?);";
```
kasunbg: You definitely fixed something in here. I'm not clear what it is though. :-) Let's see whether we can simplify this.

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
lasanthaDLPDS: Fixed 9be59e94046fe2feb8fe36c3e1d77913ba1669a7

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
lasanthaDLPDS: This is fixed 9be59e94046fe2feb8fe36c3e1d77913ba1669a7

dao/src/main/java/org/wso2/testgrid/dao/repository/TestPlanRepository.java
lasanthaDLPDS: As per the discussion we had, this is fixed with 9be59e94046fe2feb8fe36c3e1d77913ba1669a7

core/src/main/java/org/wso2/testgrid/core/command/GenerateTestPlanCommand.java
```diff
@@ -628,4 +631,35 @@ private Path getTestGridYamlLocation(String directory) {
...
         }
     }
+
+    /**
+     * Validate the testgridYaml. It must contain at least one infra provisioner, and one scenario.
+     * @param testgridYaml TestgridYaml object
+     * @return True or False, based on the validity of the testgridYaml
+     */
+    private boolean validateTestgridYaml(TestgridYaml testgridYaml) {
+        Boolean isValidTestgridYaml = true;
```
vidurananayakkara: We can optimize this code to directly return true/false instead of using a boolean variable. WDYT?

lasanthaDLPDS: Fixed with bf55b4e

PR #936, opened 2018-08-03
Add functionality to manage AWS resources

### LINKED ISSUE #793: Determine the AWS region based on the available AWS resources.
**Description:**
  In order to increase the parallelism of Testgrid jobs, we need to dynamically determine the AWS region we are going to use dynamically for each test-plan. Right now, we are limited to run on a pre-configured region (us-east-1). This is not scalable enough.                                                                                                                                             

**Affected Product Version:**
 0.9.0-m30                                                                                                                       

  N/A                                                                                                      
**Steps to reproduce:**
  The region is hard-coded into the test-plans generated via generate-testplan command. We need to determine the region at run-testplan phase instead. 
The cloudformation scripts are more or less self-contained, so this should be pretty straight-forward. Few things that I think we need factor in are: 
* AMIs are maintained per-region. We need an automated mechanism like packer to copy a given AMI into all the available regions. (Use of packer is a secondary task.)
* The EC2 ssh key-pair are maintained per-region. It's better if we can upload the same key with same name to all the regions at once.
* Some regions does not have some EC2 instance types. This is probably not an issue for us. 
                                                                                                                                        
**Related Issues:**
 Part of #528

azinneera:
**Purpose**
Introduces AWS resource management feature to allocate resources for test plans based on the availability of the regions. 
Resolves #793

**Goals**
Manage AWS resource allocation

**Approach**
- Introduces awsLimits.yaml - This file should be placed in the _$TESTGRID_HOME_. A sample file for reference is included as _core/resources/awsLimits.yaml_ 
- Before a test plan is run, AWSResourcemanger will check for an available region, allocate resources from that particular region and release them after the test run has completed.
- If the test plan for a certain deployment pattern is running for the first time, then the stack will be created in the default region specified in $TESTGRID_HOME/config.properties. The resource requirements of a deployment pattern is fetched this way. 
- Dedicated regions should be allocated for Dev and Prod environments and a default region is allocated for running the first test plans of a deployment pattern.
- The MD5 checksum of CFN is maintained with the AWSResourceRequirement to determine resource requirements. If, the hash changes, the test plan is run in the default region and the requirements are added according to the in the changed CFN
- Modified ERD
new-er

**Security checks**


common/src/main/java/org/wso2/testgrid/common/infrastructure/AWSResourceLimit.java
```diff
@@ -0,0 +1,144 @@
...
+    private int maxAllowedLimit;
+
+    @Column(name = "current_usage", nullable = false)
+    private int currentUsage;
+
+    @Transient
+    private ArrayList maxLimits;
+
+    public String getRegion() {
+
```
sameerawickramasekara: Shall we remove the unwanted newlines from all the methods

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/aws/AWSResourceManager.java
```diff
@@ -0,0 +1,297 @@
...
+            throws TestGridDAOException {
+
+        if (!awsResourceUOW.getAWSRegions().isPresent()) {
+            throw new TestGridDAOException("No AWS region information found in the database");
+        }
+        List<String> regions = awsResourceUOW.getAWSRegions().get();
+        for (String region : regions) {
+            boolean isRegionAvailable = true;
+            for (DeploymentPatternResourceUsage resourceRequirement : resourceRequirements) {
+                AWSResourceLimit awsResourceLimit = awsResourceUOW.getAvailableResource(resourceRequirement, region);
```
pasindujw: We can define AWSResourceLimit object before the for-loop and avoid creating object-per loop turn.

common/src/main/java/org/wso2/testgrid/common/DeploymentPattern.java
```diff
@@ -79,6 +88,9 @@
     @OneToMany(mappedBy = "deploymentPattern", cascade = CascadeType.ALL, orphanRemoval = true)
     private List<TestPlan> testPlans = new ArrayList<>();
 
+    @OneToMany(mappedBy = "deploymentPattern", cascade = CascadeType.ALL, orphanRemoval = true)
+    private List<DeploymentPatternResourceUsage> deploymentPatternResourceUsageList = new ArrayList<>();
```
kasunbg: A deployment pattern is a generic concept. Tying that with a resource 'usage' is not correct. We have to rethink how you have integrated your implementation into the testgrid core architecture.

kasunbg:
Where can I find the test cases? What's test coverage of this newly written code?

pom.xml
```diff
@@ -623,5 +630,6 @@
         <commons.validator.version>1.6</commons.validator.version>
         <google.gson.version>2.2.4</google.gson.version>
         <glassfish.tyrus.version>1.13.1</glassfish.tyrus.version>
+        <org.json.version>20171018</org.json.version>
```
kasunbg: Try to reuse existing json library already used.

azinneera:
Added suggested changes in commit 8da5712

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/aws/AWSRegionWaiter.java
```diff
@@ -0,0 +1,93 @@
...
+    public String waitForAvailableRegion(
+            List<AWSResourceRequirement> resourceRequirements, TimeOutBuilder timeOutBuilder)
+            throws ConditionTimeoutException, TestGridInfrastructureException, TestGridDAOException {
+
+        AWSResourceLimitUOW awsResourceLimitUOW = new AWSResourceLimitUOW();
+        String region = awsResourceLimitUOW.getAvailableRegion(resourceRequirements);
+        if (region != null) {
+            availableRegion = region;
+        } else {
+            logger.info("Waiting for an available region on AWS...");
```
kasunbg: What resource types and counts are we waiting for?

kasunbg: And, how long? The waiting time in each iteration is different since we waiting time increases exponentially (with a max ceiling).

kasunbg:
Good code. Let's this in action @dev!

PR #972, opened 2018-08-28
Change TestGrid structure to comply with Jenkins parallel pipeline execution

### LINKED ISSUE #971: Can not run parallel executors of the same job in multiple nodes
**Description:**
In TestGrid jobs we have a parallel block where the job is forked into multiple executors and then again joined after the parallel block. (We use the parallel block to execute job for each infra-combination)
Handing the parallel block to executors and other steps needed to run the parallel block are done by Jenkins. 
When the parallel blocks are running, in the memory, jenkins consider them has separate new jobs and create virtual directories (Ex: if the job name is: wso2is-intg, then for 3 parallel blocks, it will create wso2is-intg@2, @3 virtual directories). However though there are virtual directories, it seems all of them are pointing to the same workspace in the base directory. (As we don't have actual @2, @3 directories created in the disks).
This works well when we have only one node with multiple executors. But when there are more than one node, the other nodes don't have the base directory inside them. So if an executor in such node receives a parallel block to be executed, it brings an error as workspace cannot be found.

**Suggested Labels:**
Bug

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**
Execute a TestGrid job such that the parallel blocks are shared among multiple slaves.

**Related Issues:**

pasindujw:
**Purpose**
Fixes #971 

**Goals**
Spread parallel execution into multiple slaves.

**Approach**
**Isolate the parallel block working directories, so that each block run its own**
With this PR, the jobs will run as;
Sample directory;

_Node1_
```
testgridhome -> jobs ->  wso2is-intg -> workspace  (workspace used by generate-testplan command)
                                     -> 2 -> workspace   (workspaces used to run parallel block)
                                     -> 3 -> workspace
                                     -> 4 -> workspace    
                                     -> 5 -> workspace
```
_Node2_
```
testgridhome -> jobs ->  wso2is-intg -> 6 -> workspace   (workspaces used to run parallel block)
                                      -> 7 -> workspace
                                      -> 8 -> workspace
                                      -> 9 -> workspace
                                      -> 10 -> workspace
                                      -> 11 -> workspace
```

The PR include necessary changes to existing Job execution flow in order to make it self-contained.

**Security checks**

### DIFF
--- a/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
@@ -22,7 +22,7 @@ public final class S3StorageUtil {
      * @param truncated whether the truncated log or the raw log file
      * @return log file path
      */
-    public static String deriveTestRunLogFilePath(TestPlan testPlan, Boolean truncated) {
+    public static String getS3LocationForTestRunLogFile(TestPlan testPlan, Boolean truncated) {
         String productName = testPlan.getDeploymentPattern().getProduct().getName();
         String testPlanDirName = TestGridUtil.deriveTestPlanDirName(testPlan);
         String fileName = truncated ?
--- a/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
@@ -1,0 +1,48 @@
+package org.wso2.testgrid.common.util;
+
+import org.wso2.testgrid.common.TestGridConstants;
+import org.wso2.testgrid.common.TestPlan;
+import org.wso2.testgrid.common.config.ConfigurationContext;
+
+import java.nio.file.Paths;
+
+/**
+ * This Util class holds the utility methods used to manage TestGrid S3 storage .
+ *
+ * @since 1.0.0
+ */
+public final class S3StorageUtil {
+
+    /**
+     * Returns the path of the test-run log file in S3 bucket.
+     * <p>
+     * /builds/#depl_name#_#infra-uuid#_#test-run-num#/test-run.log
+     *
+     * @param testPlan test-plan
+     * @param truncated whether the truncated log or the raw log file
+     * @return log file path
+     */
+    public static String deriveTestRunLogFilePath(TestPlan testPlan, Boolean truncated) {
+        String productName = testPlan.getDeploymentPattern().getProduct().getName();
+        String testPlanDirName = TestGridUtil.deriveTestPlanDirName(testPlan);
+        String fileName = truncated ?
+                TestGridConstants.TRUNCATED_TESTRUN_LOG_FILE_NAME : TestGridConstants.TESTRUN_LOG_FILE_NAME;
+        return Paths.get(TestGridConstants.TESTGRID_JOB_DIR, productName, TestGridConstants.TESTGRID_BUILDS_DIR,
+                testPlanDirName, fileName).toString();
+    }
+
+    /**
+     * Generate the S3 bucket URL for the current environment.
+     * Ex. https://s3.amazonaws.com/bucket1
+     *
+     * @return the S3 bucket url of the environment
+     */
+    public static String getS3BucketURL() {
+        String s3BucketName = ConfigurationContext.getProperty(ConfigurationContext.
+                ConfigurationProperties.AWS_S3_BUCKET_NAME);
+        if (StringUtil.isStringNullOrEmpty(s3BucketName)) {
+            s3BucketName = TestGridConstants.AMAZON_S3_DEFAULT_BUCKET_NAME;
+        }
+        return String.join("/", TestGridConstants.AMAZON_S3_URL, s3BucketName);
+    }
+}
--- a/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
+++ b/common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
@@ -7,7 +7,7 @@ import org.wso2.testgrid.common.config.ConfigurationContext;
 import java.nio.file.Paths;
 
 /**
- * This Util class holds the utility methods used to manage TestGrid S3 storage .
+ * This Util class holds the utility methods used to manage TestGrid S3 storage.
  *
  * @since 1.0.0
  */
--- a/core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
+++ b/core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
@@ -151,17 +151,17 @@ public class RunTestPlanCommand implements Command {
     private void resolvePaths(TestPlan testPlan) {
         //infra
         Path repoPath = Paths.get(testPlan.getWorkspace(), testPlan.getInfrastructureRepository());
-        testPlan.setInfrastructureRepository(resolvePath(repoPath, testPlan));
+        testPlan.setInfrastructureRepository(resolvePath(repoPath));
         //deploy
         repoPath = Paths.get(testPlan.getWorkspace(), testPlan.getDeploymentRepository());
-        testPlan.setDeploymentRepository(resolvePath(repoPath, testPlan));
+        testPlan.setDeploymentRepository(resolvePath(repoPath));
         //scenarios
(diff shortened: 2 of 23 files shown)


common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Returns the path of the test-run log file in S3 bucket.
+     * <p>
+     * /builds/#depl_name#_#infra-uuid#_#test-run-num#/test-run.log
+     *
+     * @param testPlan test-plan
+     * @param truncated whether the truncated log or the raw log file
+     * @return log file path
+     */
+    public static String deriveTestRunLogFilePath(TestPlan testPlan, Boolean truncated) {
```
kasunbg: #deriveTestRunLogFilePath method does not belong in a class named S3Util imo..

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -131,6 +142,48 @@ public void execute() throws CommandExecutionException {
...
+     * This method resolves the absolute path to a path mentioned in
+     * the iobConfigYaml.
+     *
+     * @param path          The path to resolve
+     * @param testPlan the testPlan.yaml bean
+     * @return the resolved path
+     * @throws TestGridRuntimeException if the resolved path does not exist.
+     */
+    private String resolvePath(Path path, TestPlan testPlan) {
+        path = path.toAbsolutePath().normalize();
```
kasunbg: testPlan is an unused variable (except for the debug log).

core/src/test/java/org/wso2/testgrid/core/command/RunTestPlanCommandTest.java
```diff
@@ -130,6 +135,11 @@ public void init() throws Exception {
         testPlan.setDeploymentPattern(deploymentPattern);
         testPlan.setInfraParameters(infraParamsString);
         testPlan.setDeployerType(TestPlan.DeployerType.SHELL);
+        testPlan.setScenarioTestsRepository(workspaceDir + "/workspace/scenarioTests");
```
kasunbg: We need to use Paths API or the File.separator.

core/src/test/java/org/wso2/testgrid/core/command/RunTestPlanCommandTest.java
```diff
@@ -167,6 +177,12 @@ public void testExecute(String jobConfigFile) throws Exception {
         Assert.assertEquals(content, "Deploy server1\n", "my-deployment.txt file does not contain expected content.");
     }
 
+    private void copyWorkspaceArtifacts() throws IOException {
+        String resourceDir = "./src/test/resources/workspace";
+            FileUtils.copyDirectory(new File(resourceDir), new File(workspaceDir + "/workspace"));
+            logger.info("Copying necessary artifacts to workspace " + workspaceDir + "/workspace");
```
kasunbg: Formatting issue. We need to think about File system differences in Windows and Linux.

reporting/src/main/java/org/wso2/testgrid/reporting/TestReportEngine.java
```diff
@@ -825,6 +825,7 @@ private void postProcessSummaryTable(Map<String, InfrastructureBuildStatus> test
                 Optional<TestPlan> testPlanById = testPlanUOW.getTestPlanById(testPlanYaml.getId());
                 if (testPlanById.isPresent()) {
                     TestPlan mergedTestPlan = TestGridUtil.mergeTestPlans(testPlanYaml, testPlanById.get(), false);
+                    mergedTestPlan.setWorkspace(workspace);
```
kasunbg: Setting workspace like this in a unrelated class like TestReportEngine really complicate the matters in the long run! Shall we see a better way to fix this?

common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
pasindujw: This is referring to getting the TestRunLogFile Path(URL) in S3 bucket. I hope it is fine if the method name and method comments are changed to reflect that :) WDYT?

PS: We have the same method in TestGridUtil (Where the path is generated referring to the workspace directory) which is used when using the local storage.

reporting/src/main/java/org/wso2/testgrid/reporting/TestReportEngine.java
pasindujw: When we create a TestPlan object in the middle of the flow by calling the database or so, it is missing the workspace property. So in such times the rule would be; **whenever a TestPlan object is created, right then we need to set its workspace (which we receive at runtime via input argument)**. 
Agreed it complicates the process :) 
The ideal solution would be to keep the `workspace` in a Context where we keep run-time details of the Job. :)

common/src/main/java/org/wso2/testgrid/common/util/S3StorageUtil.java
```diff
@@ -0,0 +1,48 @@
+package org.wso2.testgrid.common.util;
+
+import org.wso2.testgrid.common.TestGridConstants;
+import org.wso2.testgrid.common.TestPlan;
+import org.wso2.testgrid.common.config.ConfigurationContext;
+
+import java.nio.file.Paths;
+
+/**
+ * This Util class holds the utility methods used to manage TestGrid S3 storage .
```
harshanL: Unnecessary empty space before '.'.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -131,6 +142,44 @@ public void execute() throws CommandExecutionException {
...
+        //keyfile
+        if (testPlan.getKeyFileLocation() != null) {
+            repoPath = Paths.get(testPlan.getWorkspace(), testPlan.getKeyFileLocation());
+            testPlan.setKeyFileLocation(resolvePath(repoPath));
+        }
+    }
+
+    /**
+     * This method resolves the absolute path to a path mentioned in
+     * the iobConfigYaml.
```
harshanL: Misspelled 'jobConfigYaml'.

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -131,6 +142,44 @@ public void execute() throws CommandExecutionException {
...
+     * @throws TestGridRuntimeException if the resolved path does not exist.
+     */
+    private String resolvePath(Path path) {
+        path = path.toAbsolutePath().normalize();
+        if (!Files.exists(path)) {
+            throw new TestGridRuntimeException("Path '" + path.toString() + "' does not exist.");
+        }
+        return path.toString();
+    }
     /**
```
harshanL: Add an empty line here.

core/src/test/java/org/wso2/testgrid/core/command/RunTestPlanCommandTest.java
```diff
@@ -107,14 +110,16 @@
     private TestPlanExecutor testPlanExecutor;
     private ScenarioExecutor scenarioExecutor;
     private String actualTestPlanFileLocation;
-
+    private String workspaceDir;
     @BeforeMethod
```
harshanL: Add an empty line here.

core/src/test/java/org/wso2/testgrid/core/command/RunTestPlanCommandTest.java
```diff
@@ -167,6 +177,13 @@ public void testExecute(String jobConfigFile) throws Exception {
         Assert.assertEquals(content, "Deploy server1\n", "my-deployment.txt file does not contain expected content.");
     }
 
+    private void copyWorkspaceArtifacts() throws IOException {
+        String resourceDir = Paths.get("./src", "test", "resources", "workspace").toString();
+        String destinationDir = Paths.get(workspaceDir, "/workspace").toString();
+            FileUtils.copyDirectory(new File(resourceDir), new File(destinationDir));
+            logger.info("Copying necessary artifacts to directory: " + destinationDir);
+    }
```
harshanL: Please reformat the above two lines.

PR #1090, opened 2018-11-23
Adding Phase2 tests

yasassri:
**Purpose**

Adding Phase2 tests

### DIFF
--- a/test/integration-tests/pom.xml
+++ b/test/integration-tests/pom.xml
@@ -61,25 +61,9 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
-        <!-- Test dependencies -->
         <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-module-testng</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-api-mockito</artifactId>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.wso2.testgrid</groupId>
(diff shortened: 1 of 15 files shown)


test/integration-tests/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.mail</groupId>
+            <artifactId>mail</artifactId>
+            <scope>test</scope>
+            <version>1.5.0-b01</version>
```
kasunbg: version need to go up.

test/integration-tests/src/test/java/org/wso2/testgrid/tests/TestPhases.java
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+    }
+
+    private void summaryTest(String jobName, String buildID) {
+
+        try {
+            testSummaryValidate(getTestPlanID(jobName, buildID), jobName);
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
```
kasunbg: log e

test/integration-tests/src/test/java/org/wso2/testgrid/tests/TestPhases.java
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+
+    private void validateLog(String testPlan) throws Exception {
+
+        String webPage = "https://testgrid-live-dev.private.wso2.com/api/test-plans/log/" + testPlan;
```
kasunbg: shall we get the hostname from a property since it is used in multiple places?

test/integration-tests/src/test/resources/email.properties
```diff
@@ -0,0 +1,6 @@
+mail.smtp.host=smtp.gmail.com
```
kasunbg: We are not reading these currently!

test/integration-tests/src/test/resources/phase1-testgrid.yaml
```diff
@@ -0,0 +1,48 @@
...
+      remoteBranch: "test"
+      description: Provision Infra for running intg tests
+      scripts:
+        - name: 'Test-Phase-1-Deployment'
+          description: ''
+          type: SHELL
+          phase: CREATE
+          file: test/integration-tests/src/test/resources/test-phase1/infra.sh
+        - name: destroy
+          file: test/integration-tests/src/test/resources/test-phase1/destroy.sh
```
kasunbg: We can test whether both these scripts were called correctly, and received the infrastructureConfig, script, and job parameters correctly.

kasunbg: Destroy script should be called at the very end. If not, that's an issue. Another thing we can assert.

test/integration-tests/src/test/resources/phase2-testgrid.yaml
```diff
@@ -0,0 +1,40 @@
...
+  containerOrchestrationEngine: None
+  parameters:
+    - JDK : ORACLE_JDK8
+  includes:
+     - CentOS
+     - mysql
+     - ORACLE_JDK8
+  provisioners:
+    - name: prod-wso2-ei-scenario-deployment
+      remoteRepository: "git@github.com:wso2/aws-ei.git"
```
kasunbg: We need a specific commit hash instead of master branch, isn't it?

test/integration-tests/pom.xml
yasassri: Fixed!

test/integration-tests/src/test/resources/email.properties
yasassri: This is bring used in the mail session object

test/integration-tests/src/test/resources/phase2-testgrid.yaml
yasassri: ATM we are using the master with a specific product version

PR #1118, opened 2018-12-17
Access Blueocean console from testplan view in webapp

sameerawickramasekara:
**Purpose**

provide a blue ocean link to the user in the test plan view

**Approach**
link is provided as an input parameter to the RunTestPlan command and then it is saved in the database.

**Security checks**

### IMPORTANT!!

- from this PR the create-or-extend-tables property is added to the persistant xml. JPA will modify the existing table in runtime and add new columns. so there is **no need of a manual database migration script.**

- Since this PR introduces Database level changes any jobs running during the PR merge will not revcieve these changes and will break during execution. Hence its **important that no builds are running until the deployments are updated.**

### DIFF
--- a/core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
+++ b/core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
@@ -74,6 +74,11 @@ public class RunTestPlanCommand implements Command {
             required = true)
     private String workspace = "";
 
+    @Option(name = "--url",
+            usage = "Jenkins URL",
+            aliases = {"-u"})
+    private String buildURL;
+
     private ProductUOW productUOW;
     private DeploymentPatternUOW deploymentPatternUOW;
     private TestPlanUOW testPlanUOW;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
```diff
@@ -74,6 +74,11 @@
             required = true)
     private String workspace = "";
 
+    @Option(name = "--url",
+            usage = "Jenkins URL",
+            aliases = {"-u"})
```
kasunbg: Make this optional?

web/src/main/react-dashboard/src/components/TestRunView.js
```diff
@@ -362,6 +364,16 @@ class TestRunView extends Component {
                   )}
           ><i className="fa fa-download" aria-hidden="true"> </i> Download Test-Run log</Button>
 
+          <Button id ="tdd" size="sm" style={{marginLeft: "10px"}}
```
kasunbg: Can't we replace existing test-run.log now? This blueocean log file makes the other one obsolete isnt it?

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
sameerawickramasekara: the default value for the required parameter is false, so this is optional. are you suggesting to make this mandatory?

web/src/main/react-dashboard/src/components/TestRunView.js
sameerawickramasekara: This change will store the blueocean urls of all the builds in the future, but for older build IMO we need to keep the existing option for a while. WDYT?

core/src/main/java/org/wso2/testgrid/core/command/RunTestPlanCommand.java
kasunbg: Just wanted to make sure this is optional. (y)

web/src/main/react-dashboard/src/components/TestRunView.js
kasunbg: +1. It is confusing to have two links for almost the same thing. How can we fix that?

PR #1176, opened 2019-03-18
Separate infra-sub-properties from infra-params

pasindujw:
**Purpose**
TestGrid infra-params has sub-properties which have been considered as another set of infra-params from one point of the code. This is conceptually wrong and this PR fixes that.

**Approach**
When those sub-properties are needed in the TG flow, TestGrid will fetch them from the database.

**Security checks**


common/src/main/java/org/wso2/testgrid/common/infrastructure/InfrastructureCombination.java
```diff
@@ -47,6 +47,17 @@ public InfrastructureCombination(Set<InfrastructureParameter> parameters) {
...
     }
 
+
+    /**
+     * Initializes an @{@link InfrastructureCombination} object with the given
+     * single infrastructure parameter.
+     *
+     * @param parameter infrastructure parameter.
+     */
+    public InfrastructureCombination(InfrastructureParameter parameter) {
```
kasunbg: I'm not sure accepting just one InfrastructureParameter by InfrastructureCombination class is good. How about we change this to accept a var-args instead (`public InfrastructureCombination(InfrastructureParameter parameter...) {
`)?

Alternatively, we can also introduce a method like #addInfrastructureParameter instead.

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -458,6 +465,19 @@ private Properties getInfraInputs(TestPlan testPlan, Script script) {
...
 
+        //Fetch infra-parms from db and include their sub-props also as properties to be passed.
+        InfrastructureParameterUOW infrastructureParameterUOW = new InfrastructureParameterUOW();
+        Properties infraParamsAsProps =  testPlan.getInfrastructureConfig().getParameters();
+        for (String key: infraParamsAsProps.stringPropertyNames()) {
+            try {
+                infrastructureParameterUOW
+                        .getInfrastructureParameter(infraParamsAsProps.getProperty(key)).get()
+                        .getProcessedSubInfrastructureParameters()
+                        .forEach(infraParam -> props.setProperty(infraParam.getType(), infraParam.getName()));
```
kasunbg: IMHO, infrastructure_parameter table is an abstract detail, and the AWSProvider should not be aware of it. If we couple these two, then it'll be, 1. hard to replace it with something else (for ex. a json file!), 2. we'll have to duplicate above logic in each and every concrete implementations of InfrastructureProvider.
 
So, I think we should always work with the TestPlan object beyond the InfrastructureCombinationProvider class.

WDYT?

common/src/main/java/org/wso2/testgrid/common/infrastructure/InfrastructureCombination.java
pasindujw: +1

infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
pasindujw: +1

dao/src/main/resources/META-INF/persistence.xml
```diff
@@ -34,7 +34,7 @@
         <properties>
             <property name="javax.persistence.jdbc.url"
                       value="jdbc:mysql://localhost:3306/testgriddb"/>
-            <property name="javax.persistence.jdbc.user" value="root"/>
+            <property name="javax.persistence.jdbc.user" value="root2"/>
```
kasunbg:
```suggestion
            <property name="javax.persistence.jdbc.user" value="root"/>
```

PR #1182, opened 2019-03-21
modified wso2-intg-test-cfn.yaml

aaujayasena:
Modified  wso2-intg-test-cfn.yaml file default WSO2InstanceType and add c3.xlarge as a valid type


aaujayasena:
This is not a valid PR. Close the PR

PR #1312, opened 2020-07-17
Support SUSE 12

NishikaDeSilva:
**Purpose**

existing ${customuserdata} is not compatible with the SUSE 12 operating system.

**Goals**

Update to support SUSE 12

### DIFF
--- a/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
+++ b/infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
@@ -739,9 +739,10 @@ public class AWSProvider implements InfrastructureProvider {
                             "SERVER=$(awk -F= '/^NAME/{print $2}' /etc/os-release)\n" +
                             "if [[ $SERVER == '\"Ubuntu\"' ]]; then\n" +
                             "update-rc.d testgrid-agent defaults\n" +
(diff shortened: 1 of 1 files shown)


infrastructure/src/main/java/org/wso2/testgrid/infrastructure/providers/AWSProvider.java
```diff
@@ -739,9 +739,10 @@ private boolean doRelease(Script script, Properties inputs, TestPlan testPlan)
                             "SERVER=$(awk -F= '/^NAME/{print $2}' /etc/os-release)\n" +
                             "if [[ $SERVER == '\"Ubuntu\"' ]]; then\n" +
                             "update-rc.d testgrid-agent defaults\n" +
-                            "elif [[ $SERVER = 'CentOS Linux' ]]; then\n" +
+                            "elif [[ $SERVER = 'CentOS Linux' ]] || [[ $SERVER = \"SLES\" ]]; then\n" +
```
msmshariq, line 742: Shouldn't this be `|| [[ $SERVER = '\"SLES\"' ]]`, can you please check?

NishikaDeSilva: With single quotes, it won't escape the double-quotes. I've tested and this works.

